PR #3, opened 2017-02-20
*: add .travis.yml

### LINKED ISSUE #2: Configure and fix Travis CI build

runcom:
Close #2 

It's worth noticing that this is just a stopgap to test this repository since lots of tests requires selinux on host and travis currently uses either ubuntu (no selinux) or containers (no selinux). The current Travis is testing just a subset of go-selinux right now...we need to find a way to use Fedora/Centos/RHEL hosts to run tests here.

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: go
+go:
+  - 1.8
+  - tip
+  - master
+
+script:
+  - go test -v -tags "selinux linux" ./...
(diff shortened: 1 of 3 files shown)


caniszczyk:
going to just merge this in since it's trivial, PullApprove will be in effect for everything afterwards

PR #4, opened 2017-02-22
Don't override system error

cpuguy83:
The error message added here provides no value as the caller already
knows all the added details. However it is covering up the underyling
system error (typically `ENOTSUP`). There is no way to handle this error before
this change.

This is the same change as opencontainers/runc#1339


runcom:
LGTM

mrunalp:
LGTM

PR #10, opened 2017-03-10
Move golang bindings to a golang subdir

rhatdan:
If we want to add container-selinux package to this repo, we need to move
the golang bindings to a golang subdir.


rhatdan:
@mrunalp @runcom PTAL
How do you get the go tests to run in the selinux/golang directory?

mrunalp:
go test /path/to/pkg
Or 
go test ./...  For all packages

crosbymichael:
Usually go package names are something like `go-selinux`, `selinux-go` or anything but having `golang` in the name ;)

runcom:
Agree with Michael

rhatdan:
Renamed to selinux/go-selinux, now should I add a Makefile to run the tests?

mrunalp:
@rhatdan :+1: to adding a Makefile for running tests.

Issue #14, opened 2017-04-24
Strong types for package API


cpuguy83:
Currently we're relying on string parsing for setting selinux labels. This requires runtime checking and is prone to error.
It would be good if the package provided a strongly typed object for interacting with.

Swarmkit uses this:

```go
type SELinuxContext struct {
	Disable bool

	User  string
	Role  string
	Type  string
	Level string
}
```

PR #15, opened 2017-04-26
WIP: *: refactor to use strong types

runcom:
Will take care of #14 

- [ ] get rid of all the proxy functions in the `label` pkg, just use `selinux` no?
- [ ] the `state` struct in `selinux` smells
- [ ] OTHERS

This is heavy WIP and I don't expect we'll finish it anytime soon, but it's to start discussing what @cpuguy83 proposed. We need to work together to ensure a pretty interface for pkg users as, right now, it's pretty awful.

@cpuguy83 @mrunalp @rhatdan PTAL

Signed-off-by: Antonio Murdaca <runcom@redhat.com>


runcom:
@cpuguy83 how would you use the new `Label` and/or `Context` struct? in `daemon/create.go`? Will `HostConfig` have a strong type for the `SecurityOpts` field?

cpuguy83:
@runcom Docker will continue to have to support the sec-opt format sadly, but at least we can parse into the new type.
With moby-project a more well structured API is on the table for sure.

rhatdan:
@runcom Could you get this rebased and we can take a look at it,

rhatdan:
Closing since this never moved forward.

PR #18, opened 2017-10-21
Reserve the MCS Level if the caller overrides the default.

rhatdan:
If a caller to label.InitLabels specifies a Level, then the library
leaks one Reserved level, and also fails to Reserve the specified level.

This could lead to a potential allocation of a duplicate level which would
eliminate the SELinux separation between two containers.

This patch eliminates the leak and correctly reserves the new level.


rhatdan:
@cpuguy83 @mrunalp @runcom PTAL

runcom:
:+1:

mrunalp:
LGTM

Issue #23, opened 2018-03-30
make getSELinuxfs() check default path first


kolyshkin:
I noticed that [libselinux](https://github.com/SELinuxProject/selinux/tree/master/libselinux) implementation tries the default path (`/sys/fs/selinux`) first (checking its fstype and readonly flag), then the old default (`/selinux`), and only if that fails it goes on to parse `/proc/self/mountinfo` in order to find the selinux mount point. The relevant commit is https://github.com/SELinuxProject/selinux/commit/e3cab998b48ab

The implementation in this repo doesn't do that, it parses /proc/self/mountinfo right away (albeit once), the code is here:
https://github.com/opencontainers/selinux/blob/76b408a8e68e301b8245ba414ab600a0d1d797d9/go-selinux/selinux.go#L93

Maybe it makes sense to amend this function to try the default first (I guess there's no need to try the old default, no one uses it now). I would create a PR, but it will mess with https://github.com/opencontainers/selinux/pull/20 so I'm filing an issue instead.

rhatdan:
Open a PR...

kolyshkin:
> Open a PR...

I had the code ready some time ago, was too lazy then to add a unit test, which I have done now. @rhatdan PTAL

kolyshkin:
Since https://github.com/opencontainers/selinux/pull/26 is now merged, this can be closed.

PR #26, opened 2018-05-03
getSELinuxfs: fix/improve/speedup plus test fixes

kolyshkin:
This is mostly to address https://github.com/opencontainers/selinux/issues/23

I would also like to simplify the assignment (either by using Go `init()` or `sync.Once()`, and get rid of `selinuxfsSet`), but haven't done that because the patch is intrusive enough already. Let me know if you want me do that, either right here (as a part of last commit) or as a followup.

### DIFF
--- a/go-selinux/selinux.go
+++ b/go-selinux/selinux.go
@@ -4,10 +4,12 @@ package selinux
 
 import (
 	"bufio"
+	"bytes"
 	"crypto/rand"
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@mrunalp @runcom PTAL

runcom:
LGTM

PR #27, opened 2018-05-21
fix selinufsMagic definition to work on 32 bit systems

rhatdan:
Need to cast selinuxfsMagic and buf.Type to uint32 so it will work
on both 64 bit and 32 bit systems

### DIFF
--- a/go-selinux/selinux.go
+++ b/go-selinux/selinux.go
@@ -106,7 +106,7 @@ func verifySELinuxfsMount(mnt string) bool {
 		}
 		return false
 	}
-	if buf.Type != selinuxfsMagic {
+	if uint32(buf.Type) != uint32(selinuxfsMagic) {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


go-selinux/selinux.go
```diff
@@ -106,7 +106,7 @@ func verifySELinuxfsMount(mnt string) bool {
 		}
 		return false
 	}
-	if buf.Type != selinuxfsMagic {
+	if uint32(buf.Type) != uint32(selinuxfsMagic) {
```
mrunalp: We could change constant definition for selinuxfsMagic to be uint32 instead of casting here.

rhatdan: That was the way the original patch was, but we sill need to cast the buf.Type,  I moved it here so it matched the C code.

mrunalp:
LGTM

PR #29, opened 2018-06-18
SELinux needs to be able to reset process label to ""

rhatdan:
When program calls SetProcessLabel("") this is the equivalent to

setexeccon(NULL) in C, it sets the label of the next exec process to
the default transition.  Currently we are ignoring this call which is
wrong.


rhatdan:
@mrunalp @runcom PTAL

mrunalp:
LGTM

runcom:
LGTM

PR #31, opened 2018-06-26
go-selinux is failing to build on non linux systems

rhatdan:
This patch makes go-selinux at least compile on non-linux systems
Basically I stubbed out all of the public content on selinux.go
for non linux systems, and copied the original selinux.go to selinux_linux.go.

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -1,0 +1,721 @@
+// +build linux
+
+package selinux
+
+import (
+	"bufio"
+	"bytes"
+	"crypto/rand"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+	"sync"
+	"syscall"
+)
+
+const (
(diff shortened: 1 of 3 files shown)


runcom:
@vbatts PTAL

rhatdan:
@runcom @mrunalp PTAL

baude:
LGTM, tested and it worked nicely

runcom:
LGTM

mrunalp:
LGTM

go-selinux/selinux_linux.go
```diff
@@ -0,0 +1,721 @@
+// +build linux
```
wking: If the idea is to keep the dummy code in this repo, I think we need a more specific tag here as well.  Linux users may have a kernel that doesn't have SELinux compiled in, or they may prefer the stubs for a particular build for some other reason.

rhatdan: SELinux code will just work without the selinux being compiled in on a linux system,  This is not an issue on non SELinux machines.

wking: > SELinux code will just work without the selinux being compiled in on a linux system, This is not an issue on non SELinux machines.

I've filed #32 with a stub tag.  Can we continue this discussion there?

PR #32, opened 2018-06-26
go-selinux: Use selinux build tag in go-selinux/

wking:
Now that we have a stub implementation for `!linux`, give Linux users access to that stub by setting the `selinux_stub` build tag.  For example, they may have a kernel without SELinux compiled in.  Or they may, for whatever reason, want a particular tool to build without real SELinux support.

The syntax is described [here][1].  In order to get the real implementation, you must be on Linux and you must not have `selinux_stub` set.

Following up on [this comment][2].

[1]: https://golang.org/pkg/go/build/#hdr-Build_Constraints
[2]: https://github.com/opencontainers/selinux/pull/31#discussion_r198182793


wking:
The current Linux implementation has some code to autodetect and gut support on Linux systems without SELinux compiled in.  For example, it returns an empty string as the SELinux mount [here][1].  And when that mount path is the empty string, the library [ignores the kernel value][2] and [returns `false`][3] for SELinux being enabled (although I don't understand [this earlier check][4]).

However, all of that will be slower for folks who know at compile time that they won't need SELinux.  And even folks who *do* have an SELinux filesystem mounted may want the stubs for a particular build.

[1]: https://github.com/opencontainers/selinux/blob/55cb7f16be130730b52cb3bdcc9351efa231130a/go-selinux/selinux_linux.go#L139
[2]: https://github.com/opencontainers/selinux/blob/55cb7f16be130730b52cb3bdcc9351efa231130a/go-selinux/selinux_linux.go#L85-L87
[3]: https://github.com/opencontainers/selinux/blob/55cb7f16be130730b52cb3bdcc9351efa231130a/go-selinux/selinux_linux.go#L83
[4]: https://github.com/opencontainers/selinux/blob/55cb7f16be130730b52cb3bdcc9351efa231130a/go-selinux/selinux_linux.go#L79-L81

rhatdan:
LGTM

wking:
I had somehow not realized that there was an existing `selinux` build tag.  I've pushed 8aad069 -> 402afa9 replacing my old `selinux_stub` with the preexisting `selinux`.

I've also added (minimal) tests for the stub implementation to `.travis.yml`.  It would be nice if the tests were based on a Go interface supplied by both the stub and real implementations, but that would be too big a change to bite off in this PR.  Still, even my very minimal tests would have caught #33 and [did catch][1] the still-in-flight #34.

[1]: https://travis-ci.org/opencontainers/selinux/jobs/397535562#L518

rhatdan:
LGTM
Although Travis is failing.

wking:
Rebased onto master now that #34 has landed.

rhatdan:
:+1:

rhatdan:
@mrunalp @runcom PTAL

rhatdan:
LGTM

runcom:
LGTM

PR #35, opened 2018-10-12
Release SELinux labels on error in init function.

rhatdan:
Currently if we have an error in the options, we will leak and allocated
SELinux label.  This change will release on error.

Also return the mountlabel, even if the user specifies to disable SELinux separation.  This keeps other containers from being able to look at this privileged
containers content, as well makeing sure all content has a label.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 go:
-  - 1.8.x
-  - 1.9.x
+  - 1.10.x
+  - 1.11.x
   - tip
   - master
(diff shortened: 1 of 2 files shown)


rhatdan:
@mrunalp @runcom PTAL

runcom:
LGTM

runcom:
@mrunalp you need a real comment LGTM and merge , I can't otherwise

mrunalp:
LGTM

PR #39, opened 2018-12-28
Verify labels when user passes them in

rhatdan:
Return errors on calls to NewContext and DupSecOpt when the user
passes in an SELinux label with less then three fields.

Currently this library will crash, when users pass in bad data.


rhatdan:
Begin fixing https://github.com/kubernetes-sigs/cri-o/issues/1987

rhatdan:
@mrunalp @runcom @vrothberg PTAL

caniszczyk:
ping @opencontainers/selinux-maintainers

mrunalp:
LGTM

PR #41, opened 2019-01-15
Expand the list of relabel exclude paths

rhatdan:
Add additional paths that we want to prevent users from accidently relabeling.

Including all paths in /, and users home directories.  Also added security
files in /etc and a couple of other common paths.


rhatdan:
@mrunalp @vrothberg @runcom PTAL

mrunalp:
LGTM

PR #46, opened 2019-03-12
Need to set process attributes not task

rhatdan:
The current setting of the kernel is failing to set the kernel keyring
within runc.  Changing to setting the process field instead of the
thread field seems to have fixed the issue.  With this change runc
is labeling the kernel keyring correctly so that it can be used within
the container.


rhatdan:
@mrunalp @eparis @vrothberg @runcom PTAL

rhatdan:
I have run this 100 times, and have not had a failure.  It seems to work but I would like @eparis to comment before merging.
@giuseppe PTAL

mrunalp:
LGTM

eparis:
If it works, I'm fine with it. The kernel won't let you set someone else's *create values.  I'm baffled why /proc/self/thread/`gittid()`/ wouldn't exist though!

PR #50, opened 2019-04-04
fix can't use empty str as label in some old kernels

lifubang:
Signed-off-by: lifubang <lifubang@acmcoder.com>

In some old kernels, for example, `3.10.0-862.el7.x86_64` in `RedHat Enterprise Linux 7.5`.
Docker in this kernel works very well.
While the docker version is:
`docker-ce=3:18.09.4-3.el7` with `containerd=1.2.5-3.1.el7`.

But runc in master version works fail:
If we use empty string as a label, it will fail:
`write /proc/self/attr/keycreate: permission denied`
It will cause `runc` fail.

If the label is not empty, it will success.

### DIFF
--- a/go-selinux/label/label_selinux.go
+++ b/go-selinux/label/label_selinux.go
@@ -101,7 +101,15 @@ func FormatMountLabel(src, mountLabel string) string {
 // SetProcessLabel takes a process label and tells the kernel to assign the
 // label to the next program executed by the current process.
 func SetProcessLabel(processLabel string) error {
+	if processLabel == "" && selinux.GetEnabled() {
+		processLabel = "unconfined_u:unconfined_r:unconfined_t:s0"
+	}
 	return selinux.SetExecLabel(processLabel)
+}
+
+// ClearProcessLabel is to clear process's label
+func ClearProcessLabel() error {
+	return selinux.SetExecLabel("unconfined_u:unconfined_r:unconfined_t:s0")
 }
 
 // SetSocketLabel takes a process label and tells the kernel to assign the
@@ -118,7 +126,15 @@ func SocketLabel() (string, error) {
 // SetKeyLabel takes a process label and tells the kernel to assign the
 // label to the next kernel keyring that gets created
 func SetKeyLabel(processLabel string) error {
+	if processLabel == "" && selinux.GetEnabled() {
+		processLabel = "unconfined_u:unconfined_r:unconfined_t:s0"
+	}
 	return selinux.SetKeyLabel(processLabel)
+}
+
+// ClearKeyLabel is to clear key label
+func ClearKeyLabel() error {
+	return selinux.SetKeyLabel("unconfined_u:unconfined_r:unconfined_t:s0")
 }
 
 // KeyLabel retrieves the current default kernel keyring label setting
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -333,11 +333,6 @@ func writeCon(fpath string, val string) error {
 	if fpath == "" {
 		return ErrEmptyPath
 	}
-	if val == "" {
-		if !GetEnabled() {
-			return nil
-		}
-	}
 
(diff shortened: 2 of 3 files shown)


lifubang:
I think `3.10.0` kernel is still used by lots of users, especially in some cloud.

go-selinux/selinux_linux.go
```diff
@@ -350,6 +345,10 @@ func writeCon(fpath string, val string) error {
 	} else {
 		_, err = out.Write(nil)
 	}
+	// for some kernels, we can't write "" as label.
+	if val == "" {
```
lifubang: @rhatdan Shall we need to check err type?

go-selinux/label/label_selinux.go
```diff
@@ -101,9 +101,17 @@ func FormatMountLabel(src, mountLabel string) string {
 // SetProcessLabel takes a process label and tells the kernel to assign the
 // label to the next program executed by the current process.
 func SetProcessLabel(processLabel string) error {
+	if processLabel == "" && selinux.GetEnabled() {
+		processLabel = "unconfined_u:unconfined_r:unconfined_t:s0"
```
lifubang: @rhatdan Does it make sense?

lifubang: Or after failure with `""`, then we try `"unconfined_u:unconfined_r:unconfined_t:s0"`?

rhatdan: No we should not hard code this label.  "" is a valid label.

rhatdan: It tells the the library to reset the kernel to use default labeling.

go-selinux/label/label_selinux.go
```diff
@@ -101,9 +101,17 @@ func FormatMountLabel(src, mountLabel string) string {
...
+	}
 	return selinux.SetExecLabel(processLabel)
 }
 
+// ClearProcessLabel is to clear process's label
+func ClearProcessLabel() error {
+	return selinux.SetExecLabel("unconfined_u:unconfined_r:unconfined_t:s0")
```
rhatdan: No we don't hard code any labels in the library.  ClearProcessLabel sets the label to "", which tells the kernel to use default labeling.

go-selinux/label/label_selinux.go
```diff
@@ -118,9 +126,17 @@ func SocketLabel() (string, error) {
 // SetKeyLabel takes a process label and tells the kernel to assign the
 // label to the next kernel keyring that gets created
 func SetKeyLabel(processLabel string) error {
+	if processLabel == "" && selinux.GetEnabled() {
+		processLabel = "unconfined_u:unconfined_r:unconfined_t:s0"
```
rhatdan: Wrong see above.

go-selinux/label/label_selinux.go
```diff
@@ -118,9 +126,17 @@ func SocketLabel() (string, error) {
...
+	}
 	return selinux.SetKeyLabel(processLabel)
 }
 
+// ClearKeyLabel is to clear key label
+func ClearKeyLabel() error {
+	return selinux.SetKeyLabel("unconfined_u:unconfined_r:unconfined_t:s0")
```
rhatdan: Wrong.

rhatdan:
The way to fix this is to do a check if /proc/self/attr/keycreate does not exist, If it does not then fail silently.
Not great, but I don't want to hard code checking of kernel versions.

lifubang:
> The way to fix this is to do a check if /proc/self/attr/keycreate does not exist, If it does not then fail silently.
> Not great, but I don't want to hard code checking of kernel versions.

OK, thanks. I will submit later.

rhatdan:
Replaced by https://github.com/opencontainers/selinux/pull/51

lifubang:
@rhatdan I think #51 can't fix this issue, please reopen this PR, I will change the code to tell you why. Thanks. Or I open a new one?

PR #51, opened 2019-04-04
Older kernels do not support keyring labeling

rhatdan:
Ignore setting the kernel keyring label if the
/proc/self/attr/keycreate file does not exists.

runc attempts to set this file on older kernels and is blowing up.


rhatdan:
@cyphar @vrothberg @mrunalp PTAL

cyphar:
This is similar to what I suggested in #49, though I do wonder why not just do it like this:

```diff
diff --git a/go-selinux/selinux_linux.go b/go-selinux/selinux_linux.go
index 51fa8de68a33..45e47f5eda00 100644
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -341,6 +341,10 @@ func writeCon(fpath string, val string) error {
 
        out, err := os.OpenFile(fpath, os.O_WRONLY, 0)
        if err != nil {
+               // Non-existent /proc files indicate a vintage kernel.
+               if os.IsNotExist(err) {
+                       err = nil
+               }
                return err
        }
        defer out.Close()
```

So any new `writeCon` users in the future will also handle old kernels.

lifubang:
@cyphar @rhatdan I think this PR can't fix my issue.
I spent some time dig into it, and with @rhatdan 's help, find the real reason.

My English is not very well, please see code in #52 .

rhatdan:
@cyphar I think the problem I have with this is it would be easy to mistake a success versus a typo in the code.
Which would not be caught until far later.  We know of a specific case of keycreate, so we handle it.

rhatdan:
@lifubang Could you show me the failure you are seeing with permission denied?  I would always want that feedback.

rhatdan:
@lifubang I updated this PR to handle your issue.

lifubang:
> @lifubang I updated this PR to handle your issue.

No, this can't fix my issue. Please see #52 .
If have any suggestions, please input in #52 . I will debug it. Thanks.

rhatdan:
What error does this patch miss?  Your patch misses Permission Denied.

lifubang:
> What error does this patch miss? Your patch misses Permission Denied.

I think we don't need it any more.

go-selinux/selinux_linux.go
```diff
@@ -406,7 +406,14 @@ func SocketLabel() (string, error) {
 // SetKeyLabel takes a process label and tells the kernel to assign the
 // label to the next kernel keyring that gets created
 func SetKeyLabel(label string) error {
-	return writeCon("/proc/self/attr/keycreate", label)
+	err := writeCon("/proc/self/attr/keycreate", label)
+	if os.IsNotExist(err) {
+		return nil
+	}
+	if label == "" && os.IsPermission(err) && !GetEnabled() {
```
lifubang: Once we really have no permission to `keycreate`? 
So I don't think we should use this err check.

lifubang:
> What error does this patch miss?
I can't explain it in one sentence because of my English's level.
Please see https://github.com/opencontainers/selinux/pull/52#issuecomment-480142795 to get full understand.

go-selinux/selinux_linux.go
rhatdan: This will check if someone executes runc on a machine with /proc/self/attr/keycreate, attempted to set "", got permission denied, and SELinux was disabled, so we ignore the error.

rhatdan:
@lifubang Lets stick to my PR.  Rather then switching back and forth.
This misses the case where the user is allowed to read keycreate but is not allowed to write to keycreate, and is writing "" with a disabled SELinux system.

go-selinux/selinux_linux.go
lifubang: Oh, not SELinux disabled, but SELinux enabled.

lifubang:
I think I need to debug it to ensure it is really right, so I will work on in my PR #52 .
If I don't explain clearly, you can see my code.

rhatdan:
@cpuguy83 Does this fix the issue you are seeing.  I will merge and open a vendor for runc if it does.  I am not in favor of the other patch.

cpuguy83:
Ok, so the issue is `GetEnabled()` is is true when runc is compiled with selinux support.

rhatdan:
@cpuguy83 Is this on an SELinux system?  Is SELinux enabled?

VERSION
```diff
@@ -1 +1 @@
-1.3.0-dev
+1.2.2
```
cyphar: Did you mean to commit this?

rhatdan: Yes I am only bumping the minor number, Since this is only a bugfix.

cpuguy83:
@rhatdan 

>  Is this on an SELinux system? Is SELinux enabled?

Yes and yes.

rhatdan:
@cpuguy83 And did a patched version of runc with this patch fix the issue.  I take it this is using an older RHEL/Centos kernel that did not support keycreate labels?

VERSION
cyphar: Sure but usually releases in OCI go through a voting process and so on. Bit odd to include it in an ordinary PR.

rhatdan: I have not been following any such procedure on the SELinux package.  I guess I could, but have not in the past.  Bottom line, I would like to get this PR Verified as fixing the issue and then vendored into runc, so we can fix it there.

cpuguy83:
@rhatdan I'm using CentOS on Azure (kernel 3.10.0-862.11.6.el7.x86_64).
Unpatched this works fine if I set `--selinux-enabled=true` on docker, but otherwise I get EPERM even with this patch.

rhatdan:
Are you seeing AVCs?  
Does it work with setenforce 0.

cpuguy83:
Denial is:

```
type=PROCTITLE msg=audit(1555611120.741:866): proctitle=2F7573722F62696E2F72756E6300696E6974
type=SYSCALL msg=audit(1555611120.741:866): arch=c000003e syscall=1 success=no exit=-13 a0=8 a1=557a6f0ab780 a2=0 a3=0 items=0 ppid=10782 pid=10790 auid=4294967295 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=4294967295 comm="runc:[2:INIT]" exe="/" subj=system_u:system_r:container_runtime_t:s0 key=(null)
type=AVC msg=audit(1555611120.741:866): avc:  denied  { create } for  pid=10790 comm="runc:[2:INIT]" scontext=system_u:system_r:container_runtime_t:s0 tcontext=system_u:object_r:unlabeled_t:s0 tclass=key
```

With `setenforce 0`  there is no issue.

rhatdan:
This is a container-selinux issue.  The latest container-selinux-2.91 and greater allows this AVC.  So give this a LGTM and we can move on.
Of course I would like to know what SELinux label is being put into the process label of a container when SELinux is disabled in docker.  It should be "", but this patch would catch that.  So I am thinking you are sending some other label.  Can you grab the runtime spec json file and see what the process label is?

@mrunalp @runcom PTAL at this PR, I believe it is correct and @cpuguy83 is experiencing an different issue, that is fixed in container-selinux package.

VERSION
cyphar: Sure, do whatever works for you.

The release procedures for all `opencontainers` projects (except I guess `selinux`) have all used the voting system we use for spec releases, but it's pretty obvious that's not a great model for non-spec projects. I'm hoping to correct this after we get `1.0.0-rc8` out.

mrunalp:
LGTM

rhatdan:
@cpuguy83 I just fixed another issue in container-selinux to allow spc_t to write unlabeled_t files.
We are packaging up the fix in container-selinux-2.100

cyphar:
Can we get a tagged release so we can bump runc? :green_heart:

rhatdan:
https://github.com/opencontainers/runc/pull/2043

PR #52, opened 2019-04-04
[WIP] fix SetKeyLabel in old kernels with selinux enabled

lifubang:
In some old kernels, for example, 3.10.0-862.el7.x86_64 in RedHat Enterprise Linux 7.5.
With `selinux enabled`.
`SetKeyLabel` with empty string `""` in the first time is not supported.
But with non empty string is supported.

It will cause runc fail if `selinuxLabel` is `""` in `config.json` in these kernels.

Signed-off-by: Lifubang <lifubang@acmcoder.com>

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -406,7 +406,21 @@ func SocketLabel() (string, error) {
 // SetKeyLabel takes a process label and tells the kernel to assign the
 // label to the next kernel keyring that gets created
 func SetKeyLabel(label string) error {
-	return writeCon("/proc/self/attr/keycreate", label)
+	if label == "" && GetEnabled() {
+		if _, err := KeyLabel(); err == io.EOF || os.IsNotExist(err) {
+			// If we got io.EOF err, this means it is the first time write "" to keycreate
+			// And in some old kernels before Linux 2.6.18, there is no keycreate file
+			// So we can ignore these errors
+			return nil
+		}
+	}
+	err := writeCon("/proc/self/attr/keycreate", label)
+	if os.IsNotExist(err) {
+		// For non empty label, In some old kernels before Linux 2.6.18,
+		// there is no keycreate file, so we can ignore it
+		return nil
+	}
+	return err
 }
 
 // KeyLabel retrieves the current kernel keyring label setting


lifubang:
I think we should not be hurry to close this PR, we need a full discussion to reduce regression s which have impact to runc.

go-selinux/label/label_selinux.go
```diff
@@ -118,7 +118,13 @@ func SocketLabel() (string, error) {
 // SetKeyLabel takes a process label and tells the kernel to assign the
 // label to the next kernel keyring that gets created
 func SetKeyLabel(processLabel string) error {
-	return selinux.SetKeyLabel(processLabel)
+	err := selinux.SetKeyLabel(processLabel)
+	if err != nil && processLabel == "" {
+		if os.IsNotExist(err) || os.IsPermission(err) {
```
rhatdan: Why would we block permission denied?

lifubang: Because in this case, the path `/proc/self/attr/keycreate` is exist, when we write to it:
(1) with empty string, it got an err: `container_linux.go:345: starting container process caused "process_linux.go:430: container init caused \"write /proc/self/attr/keycreate: permission denied\""`;
(2) with non empty string, it write success.

rhatdan: What platform is this on?  I would say this is correct behaviour.
What kernel are you running and what is the protections on /proc/self/attr/keycreate?

rhatdan: If you are running with SELinux disabled this should also not cause a write.

rhatdan: Never mind, now I understand what you are doing.

lifubang: > Never mind, now I understand what you are doing.

No, I think you may not understand what's my doing fully.
I'll write it more clearly.

lifubang: Don't worry, I will dig into it to submit a more reasonable commit.
If we are hurry to fix it in runc, I think https://github.com/opencontainers/runc/pull/2033 is safe enough.
If we are don't hurry, please make me work success on this.

lifubang:
I spent a hole morning dig into this issue. And find out in this old kernel with **`selinux enabled`**:
(1) `"/proc/self/attr/keycreate"` is always exist `since Linux 2.6.18`;
(2) When we write `""` to `"/proc/self/attr/keycreate"` in the first time, it will cause `write /proc/self/attr/keycreate: permission denied` error;
(3) When we read `"/proc/self/attr/keycreate"` in the first time, it will get an error `io.EOF`;
(4) After we write `non empty` string to `"/proc/self/attr/keycreate"` , then we can read and write to `"/proc/self/attr/keycreate"` in any time without any errors.

So I think we can use `io.EOF` to check we should write `""` to `"/proc/self/attr/keycreate"`.

I have submit the latest commit. PTAL @cyphar @rhatdan Thanks.
If have any suggestions, please input here. I will debug it.

go-selinux/label/label_selinux.go
lifubang: I have remove `os.IsPermission`. Please see https://github.com/opencontainers/selinux/pull/52#issuecomment-480142795.

go-selinux/label/label_selinux.go
```diff
@@ -118,7 +119,19 @@ func SocketLabel() (string, error) {
...
-	return selinux.SetKeyLabel(processLabel)
+	if processLabel == "" && selinux.GetEnabled() {
+		if _, err := KeyLabel(); err == io.EOF {
+			// If we got io.EOF err, this means it is the first time write "" to keycreate
+			// So we can ignore it
+			return nil
+		}
+	}
+	err := selinux.SetKeyLabel(processLabel)
+	if err != nil && processLabel == "" && os.IsNotExist(err) {
```
cyphar: I don't think `processLabel == ""` should be here. really it should just be:

```
if os.IsNotExist(err) {
}
```

With the note that `os.IsNotExist(nil) == false`.

go-selinux/label/label_selinux.go
```diff
@@ -118,7 +119,19 @@ func SocketLabel() (string, error) {
...
 func SetKeyLabel(processLabel string) error {
-	return selinux.SetKeyLabel(processLabel)
+	if processLabel == "" && selinux.GetEnabled() {
+		if _, err := KeyLabel(); err == io.EOF {
+			// If we got io.EOF err, this means it is the first time write "" to keycreate
+			// So we can ignore it
+			return nil
+		}
+	}
+	err := selinux.SetKeyLabel(processLabel)
```
cyphar: Why not put this check inside `selinux.SetKeyLabel`? Seems a bit odd to put it in this wrapper.

lifubang: Yes, it is.
If the user just use `"github.com/opencontainers/selinux/go-selinux"`, not `"github.com/opencontainers/selinux/go-selinux/label"`. 
I have update it.

go-selinux/label/label_selinux.go
lifubang: Thanks for your review. I have update it.

go-selinux/selinux_linux.go
```diff
@@ -406,7 +406,19 @@ func SocketLabel() (string, error) {
 // SetKeyLabel takes a process label and tells the kernel to assign the
 // label to the next kernel keyring that gets created
 func SetKeyLabel(label string) error {
-	return writeCon("/proc/self/attr/keycreate", label)
+	if label == "" && GetEnabled() {
+		if _, err := KeyLabel(); err == io.EOF {
```
rhatdan: Why not check ENoExists here?

lifubang: Thanks, Because we add it after write action. Anyway, I think we can add. I will add it later.

lifubang: And the ENoExists check after `writeCon` should also be remained.

lifubang: Thank you for your review, it will reduce one file write call, does these code change make sense?

rhatdan: It will still fail on Permission Denied.  If the user tries to write to a KeyLabel they are allowed to read but not allowed to write.
I am not even sure what the io.EOF does?

lifubang: I think it does make sense that Permission Denied err when write to keycreate should be throwed to runc.

lifubang: > I am not even sure what the io.EOF does?

I think it is what is the older kernel really mean! It also be very very strange to me, but it really happens. I think this is a selinux bug in this old kernel.

rhatdan: What OS and What Kernel?

lifubang: > What OS and What Kernel?

I think I have provided it in PR description.
In some old kernels, for example, 3.10.0-862.el7.x86_64 in RedHat Enterprise Linux 7.5.
With selinux enabled.

lifubang:
@rhatdan Could you tell me when you submit the PR https://github.com/opencontainers/runc/pull/2012, you test it in what OS version and what kernel version?

rhatdan:
I tested it in Fedora.

cpuguy83:
Note there is also a regression with just the default docker setup (where `--selinux-enabled=false`), runc is unable to create containers as it gets EPERM when trying to SetKeyLabel.

rhatdan:
The other patch got merged.

kolyshkin:
> The other patch got merged.

For the sake of other lost souls reading this, "the other patch" here means https://github.com/opencontainers/selinux/pull/49

PR #53, opened 2019-05-30
Add PeerLabel() and SetTaskLabel()

jbrindle:
Add PeerLabel, which is equivalent to getpeercon()
Add SetTaskLabel, which is equivalent to setexeccon()


rhatdan:
That did not last long...

jbrindle:
it took a few tries to make travis-ci happy :) there should be an open PR.

PR #55, opened 2019-05-30
Add SetTaskLabel() and PeerLabel()

jbrindle:
Add SetTaskLabel, which is equivalent to setexeccon()
Add PeerLabel, which is equivalent to getpeercon()

Signed-off-by: Joshua Brindle <joshua.brindle@crunchydata.com>

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+	"golang.org/x/sys/unix"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


rhatdan:
You need the following for the SELinux disabled systems.

func SetTaskLabel(label string) error {
	return nil
}

Any chance you can add a test for this.

jbrindle:
I missed the comment about the tests. I looked at the test stuff when I did this and 1) SetTaskLabel is trivial to test but requires dyntransition to some other domain which is completely policy dependent and 2) PeerLabel could be testable by creating a client/server connection but there will be no answer if netlabelctl hasn't already been used to set up maps and doi's. netlabelctl isn't installed by default either, and requires root to set up the labeling.

I can make tests that require all of the preconditions be set up if you want but it'll take some time.

rhatdan:
Can't you just copy the setexeccon code now.  The tests for this should be the same.  My concern is not that SELinux policy gets screwed up, it is that we screw up compiling on systems without SELinux enabled.

jbrindle:
Do you mean just adding this to selinux_stub_test.go?

	if err := SetTaskLabel("foobar"); err != nil {
		t.Fatal(err)
	}
	if _,err := PeerLabel(0); err != nil {
		t.Fatal(err)
	}

If that is what you are asking for then yea, no problem at all to add that.

rhatdan:
Yes.

rhatdan:
LGTM
@mrunalp @runcom PTAL

PR #59, opened 2019-09-27
selinux: verify that writes to /proc/... are on procfs

cyphar:
This is an additional mitigation for CVE-2019-16884. The primary problem
is that Docker can be coerced into bind-mounting a file system on top of
/proc (resulting in label-related writes to /proc no longer happening).

While runc is working on mitigations against permitting the mounts, this
helps avoid go-selinux from being tricked into writing to non-procfs
files. This is not a perfect solution (after all, there might be a
bind-mount of a different procfs file over the target) but in order to
exploit that you would need to be able to tweak a config.json pretty
specifically (which thankfully Docker doesn't allow).

See opencontainers/runc#2128

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@mrunalp @giuseppe @vrothberg PTAL

cyphar:
I changed the error message formatting to match the updated runc PR.

/cc @opencontainers/selinux-maintainers

mrunalp:
LGTM

PR #63, opened 2020-02-14
Add KVMProcessLabel to identify the SELinux label for container vms

rhatdan:
Containers running inside of VM's will need to run with a VM Label
rather then a container label, since they need different access then
containers.

Adding interface to allow Kata containers to run from container runtimes
that support SELinux.


rhatdan:
If the container engines know that they are going to use a kvm based container they need to set the SELinux label to the KVM (virutualizationDomainLabel) container label rather then the LXC/container label.

rhatdan:
@mrunalp @vrothberg @mheon PTAL

rhatdan:
We could log add logging of errors, but in lot of cases like the file not being there, is expected on SELinux disabled or non supported platforms.  So returning "" indicates that the system should not use labelling.

rhatdan:
We should probably wait on this though, since we might need a different label for kvm separated containers then svirt_t.

And I might need to add that the the lxc_contexts file.

rhatdan:
Testing with https://github.com/containers/libpod/pull/5690

rhatdan:
@vrothberg @kolyshkin @mrunalp @giuseppe PTAL

go-selinux/selinux_linux.go
```diff
@@ -726,26 +730,64 @@ func ContainerLabels() (processLabel string, fileLabel string) {
...
+	}
+
+	return addMcs(processLabel, labels["file"])
+}
+
+/*
+InitContainerLabels returns the default processLabel and file labels to be
+used for containers running an init system like systemd by the calling process.
+*/
+func InitContainerLabels() (string, string) {
```
mrunalp, line 757: If I understand correctly this is going to be used for labeling containers that start systemd? If so, can we pull this into a separate commit or a PR?

rhatdan:
@mrunalp PTAL

PR #66, opened 2020-02-19
Use /proc/thread-self if available

kolyshkin:
As suggested by @stephensmalley in https://github.com/opencontainers/selinux/pull/64#issuecomment-588264641,
Linux >= 3.17 provides `/proc/thread-self` symlink so there's no need to use `gettid()`.

Let's use this. If not available, fall back to the older method.


mrunalp:
LGTM

rhatdan:
LGTM

PR #67, opened 2020-02-19
xattr: use x/sys/unix, simplify

kolyshkin:
1. For quite some time `Lgetxattr`/`Lsetxattr` are available in `x/sys/unix`, so let's use those.

2. Simplify the `ERANGE` handling logic.


mrunalp:
LGTM

rhatdan:
LGTM

PR #69, opened 2020-02-21
Translation support and CreateContext

jbrindle:


go-selinux/selinux_linux.go
```diff
@@ -410,6 +427,13 @@ func CanonicalizeContext(val string) (string, error) {
 	return readWriteCon(filepath.Join(getSelinuxMountPoint(), "context"), val)
 }
 
+/*
+CreateContext requests the type transition from source to target for class  from the kernel.
```
rhatdan: This seems badly named.

go-selinux/go.mod
```diff
@@ -0,0 +1,7 @@
+module github.com/yulicrunchy/selinux-go/go-selinux
```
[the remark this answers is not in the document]
jbrindle: yea, accidental PR, will clean up. Thanks

go-selinux/selinux_linux.go
jbrindle: What name would you prefer? ComputeCreateContext?

PR #70, opened 2020-02-26
Complete syscall -> x/sys/unix switch, fix lgetxattr corner case

kolyshkin:
These two are really unrelated but since they are both very simple I combined both to this single PR.

1. Complete the switch to `x/sys/unix`, and remove own ST_RDONLY const.

2. Fix a corner case in lgetxattr.

    lgetxattr(2) man page says:
    
    > If size is specified as zero, these calls return the  current  size  of
    > the  named extended attribute (and leave value unchanged).  This can be
    > used to determine the size of the buffer that should be supplied  in  a
    > subsequent  call.   (But, bear in mind that there is a possibility that
    > the attribute value may change between the two calls,  so  that  it  is
    > still necessary to check the return status from the second call.)
    
    The current code does not handle the "But..." case, i.e. when the size
    changes between the two calls, and the new size is larger.
    
    Fix that.

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -17,7 +17,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"syscall"
 
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@mrunalp @giuseppe PTAL

rhatdan:
@saschagrunert PTAL

mrunalp:
LGTM

PR #71, opened 2020-02-27
Add uint32 cast to unix.SELINUX_MAGIC

saschagrunert:
Otherwise this value will be treated as `int32`, which results in an
build error in 386 environments:

```
selinux_linux.go:118:14: constant 4185718668 overflows int32
```

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -115,7 +115,7 @@ func verifySELinuxfsMount(mnt string) bool {
 		return false
 	}
 
-	if buf.Type != unix.SELINUX_MAGIC {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@mrunalp @giuseppe PTAL

mrunalp:
LGTM

kolyshkin:
Thanks for catching this! Looks like we need to have some kind of CI for this repo

saschagrunert:
Yeah some cross compilation builds would be good. I think I can follow up on that :)

thaJeztah:
@rhatdan is it possible to tag a v1.3.3 with this fix? (this regression was included in the latest v1.3.2 patch)

PR #74, opened 2020-02-27
Add support for ComputeCreateContext and ClassIndex

jbrindle:
ComputeCreateContext is the same as compute_create in libselinux and takes
source context, target context, and class index and returns
the computed transition context. The class index can be retrieved
with ClassIndex().

Signed-off-by: Joshua Brindle <joshua.brindle@crunchydata.com>

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -280,6 +280,23 @@ func readCon(fpath string) (string, error) {
 	return strings.Trim(retval, "\x00"), nil
 }
 
+// ClassIndex returns the int index for an object class in the loaded policy, or -1 and an error
+func ClassIndex(class string) (int, error) {
+	permpath := fmt.Sprintf("class/%s/index", class)
+	indexpath := filepath.Join(getSelinuxMountPoint(), permpath)
+
+	indexB, err := ioutil.ReadFile(indexpath)
+	if err != nil {
+		return -1, err
+	}
+	index, err := strconv.Atoi(string(indexB))
+	if err != nil {
+		return -1, err
+	}
+
+	return index, nil
+}
+
 // SetFileLabel sets the SELinux label for this path or returns an error.
 func SetFileLabel(fpath string, label string) error {
 	if fpath == "" {
@@ -406,6 +423,18 @@ can be used to see if two contexts are equivalent
 */
 func CanonicalizeContext(val string) (string, error) {
 	return readWriteCon(filepath.Join(getSelinuxMountPoint(), "context"), val)
+}
+
+/*
+ComputeCreateContext requests the type transition from source to target for class  from the kernel.
+*/
+func ComputeCreateContext(source string, target string, class string) (string, error) {
+	classidx, err := ClassIndex(class)
+	if err != nil {
+		return "", err
+	}
+
+	return readWriteCon(filepath.Join(getSelinuxMountPoint(), "create"), fmt.Sprintf("%s %s %d", source, target, classidx))
 }
 
 func readWriteCon(fpath string, val string) (string, error) {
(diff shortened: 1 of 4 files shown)


go-selinux/selinux_linux.go
```diff
@@ -282,6 +282,23 @@ func readCon(fpath string) (string, error) {
 	return strings.Trim(retval, "\x00"), nil
 }
 
+func ClassIndex(class string) int {
```
kolyshkin: Please add documentation

kolyshkin: The usual go convention is to return a value and an error. Returning -1 as an indication of an error is way too C-like, and we lose the error information.

go-selinux/selinux_linux.go
```diff
@@ -282,6 +282,23 @@ func readCon(fpath string) (string, error) {
...
+	permpath := fmt.Sprintf("class/%s/index", class)
+	indexpath := filepath.Join(getSelinuxMountPoint(), permpath)
+
+	in, err := os.Open(indexpath)
```
kolyshkin: Not a strong preference, but maybe `ioutil.ReadFile()` and `strconv.Atoi()` will look better?

go-selinux/selinux_linux_test.go
```diff
@@ -200,3 +200,27 @@ func TestSecurityCheckContext(t *testing.T) {
...
+
+func TestComputeCreateContext(t *testing.T) {
+	if !GetEnabled() {
+		t.Skip("SELinux not enabled, skipping.")
+	}
+
+	// This may or may not be in the loaded policy but any refpolicy based policy should have it
+	t.Logf("ComputeCreateContext(\"system_u:system_r:init_t:s0\", \"system_u:object_r:tmp_t:s0\", ClassIndex(\"file\"))")
+	context, err := ComputeCreateContext("system_u:system_r:init_t:s0", "system_u:object_r:tmp_t:s0", ClassIndex("file"))
+	if err != nil {
```
kolyshkin: missing error check from `ClassIndex()`

go-selinux/selinux_linux.go
```diff
@@ -410,6 +427,13 @@ func CanonicalizeContext(val string) (string, error) {
 	return readWriteCon(filepath.Join(getSelinuxMountPoint(), "context"), val)
 }
 
+/*
+ComputeCreateContext requests the type transition from source to target for class  from the kernel.
+*/
+func ComputeCreateContext(source string, target string, class int) (string, error) {
```
kolyshkin: I'd prefer a bit higher-level interface -- pass the class as a string (e.g. "file") and do a conversion to index internally. It's easy to do it with caching and map.

go-selinux/selinux_linux.go
```diff
@@ -282,6 +282,23 @@ func readCon(fpath string) (string, error) {
...
+
+	indexB, err := ioutil.ReadFile(indexpath)
+	if err != nil {
+		return -1, err
+	}
+	index, err := strconv.Atoi(string(indexB))
+	if err != nil {
+		return -1, err
+	}
+
```
kolyshkin: Perhaps we can cache the result in a map, so we won't have to reread it every time?

I'm not sure how often it is supposed to be called.

rhatdan: One issue here would be in a long running process, the selinux policy could change and the results could change.  This should not be called enough where caching would matter.

rhatdan:
@jbrindle Tests are failing.

go-selinux/selinux_linux.go
jbrindle: If we start doing a full avc implementation we will have to add caching but at that point we'll have a netlink socket so we'll know when the policy changes and can flush everything. Until then @rhatdan is correct, a long running process could possibly cache stale values

jbrindle:
Sorry about the tests, I forgot to run the non-selinux version. Should be fixed now.

rhatdan:
LGTM
@kolyshkin @mrunalp @giuseppe PTAL

PR #75, opened 2020-02-28
Add cross build to travis

saschagrunert:
This adds cross compilation targets to travis to avoid issues when
building for different platforms.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,30 @@
 language: go
 go:
-  - 1.10.x
-  - 1.11.x
+  - 1.12.x
+  - 1.13.x
   - tip
-  - master
-
-env:
-  global:
-    - BUILDTAGS="selinux"
 
 before_install:
   - go get -u golang.org/x/lint/golint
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,19 +1,28 @@
...
 go:
-  - 1.10.x
-  - 1.11.x
-  - tip
-  - master
-
-env:
-  global:
-    - BUILDTAGS="selinux"
+  - 1.12.x
```
saschagrunert: Let's move forward with go 1.12

rhatdan: Should we test on 1.13 and tip?

rhatdan:
LGTM other then question on versions.

.travis.yml
saschagrunert: Okay, I add those versions as well.

rhatdan:
@kolyshkin @mrunalp @giuseppe PTAL

mrunalp:
LGTM

PR #76, opened 2020-02-28
Use golangci-lint for linting

saschagrunert:
Use golangci-lint for the CI linting step as well as providing a config
which enables all succeeding linters for now.


rhatdan:
@saschagrunert Move the testing to go1.12 and beyond.  We no longer need to test 1.11 and 1.10.

saschagrunert:
> @saschagrunert Move the testing to go1.12 and beyond. We no longer need to test 1.11 and 1.10.

Alright, changed it there: https://github.com/opencontainers/selinux/pull/75#discussion_r385647075

saschagrunert:
Merged the master via the web interface.

kolyshkin:
> Merged the master via the web interface.

It would be better to rebase this on top of the current master rather than merge the master to here. Now we have that extra merge commit.

Makefile
```diff
@@ -12,13 +16,18 @@ test: check-gopath
 	go test -timeout 3m -tags "${BUILDTAGS}" ${TESTFLAGS} -v ./...
 	go test -timeout 3m ${TESTFLAGS} -v ./...
 
+${GOLANGCI_LINT}:
+	export \
+		VERSION=v1.23.6 \
```
kolyshkin: 1. `1.23.7`
2. I think `golangci-lint` installation belongs to `.travis.yml` rather than Makefile. For example, I do have it installed on my system already, and won't need an extra copy for this package when running `make lint`.

saschagrunert: > 1. `1.23.7`

Updated it to the latest version.

> 2. I think `golangci-lint` installation belongs to `.travis.yml` rather than Makefile. For example, I do have it installed on my system already, and won't need an extra copy for this package when running `make lint`.

I disagree, because this would lead into inconsistent results depending on the machine we're running `make lint`. I would prefer in always running the same version of golangci-lint, independently if on CI or locally. WDYT?

saschagrunert:
> > Merged the master via the web interface.
> 
> It would be better to rebase this on top of the current master rather than merge the master to here. Now we have that extra merge commit.

Rebased on top of the latest master branch

.golangci.yml
```diff
@@ -0,0 +1,124 @@
+---
+run:
+  build-tags:
+    - selinux
+  concurrency: 6
+  deadline: 5m
+linters:
+  disable-all: true
+  enable:
```
vrothberg: I prefer to invert the logic. Enable all by default and disable selectively. It's easier to maintain as we fetch new linters on updates and we actually know which linters are not working.

saschagrunert: Yep, changed it to an inverting logic.

vrothberg: Awesome, thanks!

rhatdan:
LGTM

PR #78, opened 2020-02-28
use pkg/errors for all errors

thaJeztah:
use errors.Wrapf() to preserve the original error where possible.

### DIFF
--- a/go-selinux/label/label_selinux.go
+++ b/go-selinux/label/label_selinux.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/opencontainers/selinux/go-selinux"
+	"github.com/pkg/errors"
 )
 
 // Valid Label Options
@@ -21,7 +22,7 @@ var validOptions = map[string]bool{
 	"level":    true,
 }
 
-var ErrIncompatibleLabel = fmt.Errorf("Bad SELinux option z and Z can not be used together")
+var ErrIncompatibleLabel = errors.New("Bad SELinux option z and Z can not be used together")
 
 // InitLabels returns the process label and file labels to be used within
 // the container.  A list of options can be passed into this function to alter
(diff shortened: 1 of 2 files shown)


kolyshkin:
@thaJeztah can you please amend the patch by converting all uses of `fmt.Errorf` to `errors.Wrapf()` (there are some more in `go-selinux/label/`)?

go-selinux/label/label_selinux.go
```diff
@@ -52,11 +53,11 @@ func InitLabels(options []string) (plabel string, mlabel string, Err error) {
 				return "", mountLabel, nil
 			}
 			if i := strings.Index(opt, ":"); i == -1 {
-				return "", "", fmt.Errorf("Bad label option %q, valid options 'disable' or \n'user, role, level, type, filetype' followed by ':' and a value", opt)
+				return "", "", errors.Errorf("Bad label option %q, valid options 'disable' or \n'user, role, level, type, filetype' followed by ':' and a value", opt)
```
thaJeztah: Wondering; should this also check if there's anything _after_ the colon? (e.g., `disable:` is currently "valid")

thaJeztah:
@kolyshkin updated the "label" package to also use pkg/errors

rhatdan:
LGTM

go-selinux/label/label_selinux.go
kolyshkin: Yeah, this is not a very good check; should be addressed separately

rhatdan: @kolyshkin Do you approve of this patch?

rhatdan:
@mrunalp @giuseppe @vrothberg @kolyshkin PTAL

kolyshkin:
@thaJeztah and now I think perhaps we should have used new `%w` directive of `fmt.Errorf()` instead.

PR #81, opened 2020-03-02
deprecate selinux aliases, and refactor "stubs" to use selinux stubs

thaJeztah:
Reading the source code, I _think_ the wrappers/aliases in the "label" package were intended as a transition to use the selinux package, some of the wrappers providing some additional code for backward compatibility.

With many of these wrappers now being a direct alias for their equivalent in the "selinux", it would appear to make sense to _deprecate_ them, and encourage users to use the selinux package.

Some of the stubs in the labels package already had an equivalent stub in the selinux package, so for those it was no longer needed to repeat "stubbing" the functions, so for those, the alias could be used "unconditionally".

After this change, some stubs remain (some that conditionally call code, based on `selinux.GetEnabled()`), but given that `selinux.GetEnabled()` in itself is also "stubbed" in the `selinux` package, those _could_ also be removed (perhaps in a follow-up).

@kolyshkin @rhatdan PTAL if my assumptions were correct (regarding transition/deprecation). Given that some files were renamed, and things were moved around, this PR is likely easiest to review by reviewing individual commits.


thaJeztah:
Ahum; looks like I need to review the build-tags in these files;

```
$ make
GOOS=linux GOARCH=amd64 go build -tags selinux ./...
GOOS=linux GOARCH=386 go build -tags selinux ./...
GOOS=linux GOARCH=arm go build -tags selinux ./...
GOOS=linux GOARCH=arm64 go build -tags selinux ./...
GOOS=linux GOARCH=ppc64le go build -tags selinux ./...
GOOS=linux GOARCH=s390x go build -tags selinux ./...
GOOS=windows GOARCH=amd64 go build -tags selinux ./...
go-selinux/label/label.go:4:2: build constraints exclude all Go files in /home/travis/gopath/src/github.com/opencontainers/selinux/go-selinux
Makefile:17: recipe for target 'build-cross' failed
make: *** [build-cross] Error 1
```

thaJeztah:
Fixed: issue was because the `selinux` package didn't allow the stubs to be used for non-linux, whereas the `label` package did. Added an extra commit to allow the selinux package (stubs) to be used on non-linux.

go-selinux/label/label_stub.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func Relabel(path string, fileLabel string, shared bool) error {
+	return nil
+}
+
+// DisableSecOpt returns a security opt that can disable labeling
+// support for future container processes
+func DisableSecOpt() []string {
+	// TODO the selinux.DisableSecOpt stub returns []string{"disable"} instead of "nil"
```
thaJeztah: Added a TODO for this; wasn't sure if changing the return would cause compatibility issues for existing uses of these stubs.

rhatdan:
The original goal, many years ago was to allow other labelling systems like Smack to work.  Not sure if this change would break that possibility or not.  But since no one has ever looked into adding alternative labelling, then I guess this is ok.

go-selinux/label/label_selinux.go
```diff
@@ -75,11 +75,8 @@ func InitLabels(options []string) (plabel string, mlabel string, Err error) {
 	return processLabel, mountLabel, nil
 }
 
-func ROMountLabel() string {
-	return selinux.ROFileLabel()
-}
-
-// DEPRECATED: The GenLabels function is only to be used during the transition to the official API.
+// Deprecated: The GenLabels function is only to be used during the transition
```
rhatdan: This has been depracated for a long time, why not just drop it.

kolyshkin:
/lgtm

Thank you @thaJeztah, this looks awesome! I guess we can remove the deprecated stuff in the next major release.

thaJeztah:
yes, if we want to accommodate go modules users, we can't really remove things unless we do a major update

well, or rename the package like they did with the protobuf 😬

rhatdan:
@thaJeztah Remove the [rfc] and we can merge.

thaJeztah:
done 👍

rhatdan:
LGTM

PR #82, opened 2020-03-02
make sure the underlying error is checked

thaJeztah:
Some of the errors were wrapped; make sure we unwrap the error before checking the cause.

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -495,10 +495,10 @@ func PeerLabel(fd uintptr) (string, error) {
 // label to the next kernel keyring that gets created
 func SetKeyLabel(label string) error {
 	err := writeCon("/proc/self/attr/keycreate", label)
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

thaJeztah:
Thanks for reviewing; yes, not all paths seemed critical, but getting the underlying error didn't seem to hurt 👍

rhatdan:
One thing we might want to look into is makeing the recursive chcon multi-threaded.  We have seem some serious improvements in speed doing this with chown.

thaJeztah:
Perhaps there should be a tracking issue for that, in case someone wants to pick up the work to do that

rhatdan:
https://github.com/opencontainers/selinux/issues/84
I also added some code to show how we might want to do it.  Of course, if someone better then me in golang wants to improve it and run with it, I am willing to accept a PR.

Issue #84, opened 2020-03-03
Make recursive chcon threaded.


rhatdan:
Currently chcon of large directories takes a very long time.  We have experimented with kicking off a thread for each directory and recusively calling chown, and have seen serious improvements in speed.
Up to 10 times faster

Here is my experimental code
https://github.com/rhatdan/tools/blob/master/chown.go

It would be nice to use this inside of the selinux library.

kolyshkin:
Played a bit with it (before seeing @rhatdan code), didn't get 10x speedup but 3x is not bad, too, I guess. Please see #85.

kolyshkin:
Fixed by https://github.com/opencontainers/selinux/pull/89

PR #85, opened 2020-03-03
[WIP/DNM] Chcon speedup

kolyshkin:
Fixes: https://github.com/opencontainers/selinux/issues/84

Improve `Chcon()` performance about 3x by using:
 - [github.com/karrick/godirwalk](https://github.com/karrick/godirwalk/) to avoid calling `stat()` on every file;
 - parallel Walk (based on [github.com/stretchr/powerwalk](https://github.com/stretchr/powerwalk/)).

A benchmark is included.

Some benchmarks (a working set is linux kernel source tree with about 70K files):

| Implementation | Time, seconds |
| --- | ---- |
| original (before this PR) | 0.42 |
| `find(3)` | 0.11 |
| `chcon(3)` | 0.35 |
| `karrick/godirwalk` | 0.33 |
| this PR | 0.16 |

NOTE: `find(3)` time is for reference only.

### Testing notes

```
     cd go-selinux
     # create lots of files in this directory; I'm using kernel src
     cp -a ~/git/linux .
     # check we have lots of files and get a baseline fron find(3)
     time find . -print | wc -l
     # get another baseline from chcon(3)
     time chcon -R --reference selinux_linux_test.go .
     # compile the test binary; this is important
     go test -c -tags selinux -bench .
     # run the benchmark
     ./go-selinux.test -test.bench .
```

### DIFF
--- a/go-selinux/pwalk.go
+++ b/go-selinux/pwalk.go
@@ -8,116 +8,83 @@ package selinux
  */
 
 import (
+	"runtime"
 	"sync"
 
 	"github.com/karrick/godirwalk"
 	"github.com/pkg/errors"
 )
 
-// DefaultConcurrentWalks is the default number of files that will be walked at the
-// same time when the Walk function is called.
-// To use a value other than this one, use the WalkLimit function.
-const DefaultConcurrentWalks int = 100
-
-// Walk walks the file tree rooted at root, calling walkFn for each file or
-// directory in the tree, including root. All errors that arise visiting files
-// and directories are filtered by walkFn. The output is non-deterministic.
-// WalkLimit does not follow symbolic links.
+// Walk walks the file tree rooted at root, calling options.Callback
+// for each file or directory in the tree, including root. All errors
+// that arise visiting files and directories are filtered by Callback.
+// The output is non-deterministic.
 //
 // For each file and directory encountered, Walk will trigger a new Go routine
-// allowing you to handle each item concurrently.  A maximum of DefaultConcurrentWalks
-// walkFns will be called at any one time.
+// allowing you to handle each item concurrently. A maximum of twice the
+// runtime.NumCPU() Callback goroutines will be called at any one time.
 func Walk(root string, options *godirwalk.Options) error {
-	return WalkLimit(root, options, DefaultConcurrentWalks)
+	return WalkLimit(root, options, runtime.NumCPU()*2)
 }
 
-// WalkLimit walks the file tree rooted at root, calling walkFn for each file or
-// directory in the tree, including root. All errors that arise visiting files
-// and directories are filtered by walkFn. The output is non-deterministic.
-// WalkLimit does not follow symbolic links.
+// WalkLimit walks the file tree rooted at root, calling options.Callback
+// for each file or directory in the tree, including root. All errors
+// that arise visiting files and directories are filtered by Callback.
+// The output is non-deterministic.
 //
 // For each file and directory encountered, Walk will trigger a new Go routine
-// allowing you to handle each item concurrently.  A maximum of limit walkFns will
-// be called at any one time.
+// allowing you to handle each item concurrently. A maximum of limit
+// Callback goroutines will be called at any one time.
 func WalkLimit(root string, options *godirwalk.Options, limit int) error {
 	// make sure limit is sensible
 	if limit < 1 {
-		panic("powerwalk: limit must be greater than zero.")
+		return errors.Errorf("walk(%q): limit must be greater than zero", root)
 	}
 
-	// filesMg is a wait group that waits for all files to
-	// be processed before finishing.
-	var filesWg sync.WaitGroup
+	files := make(chan *walkArgs, 128)
+	errCh := make(chan error, 1) // get the first error, ignore others
+	// save the original callback func
+	callback := options.Callback
 
-	// files is a channel that receives lists of channels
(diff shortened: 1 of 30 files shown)


kolyshkin:
Let me fix the linting issues (coming from stretchr/powerwalk code)

rhatdan:
@rhvgoyal FYI

rhatdan:
Nice,  @kolyshkin 

I am not crazy about adding the other dependencies, but if this helps speed this up, I guess we can take it.

LGTM

The speed advantage depends on the number of directories that you spread out your content into.

Parallel Chown discussion happening here.
https://bugzilla.redhat.com/show_bug.cgi?id=1773211

rhatdan:
@giuseppe @vrothberg PTAL

rhatdan:
@mrunalp PTAL

go-selinux/pwalk.go
```diff
@@ -0,0 +1,128 @@
...
+// and directories are filtered by walkFn. The output is non-deterministic.
+// WalkLimit does not follow symbolic links.
+//
+// For each file and directory encountered, Walk will trigger a new Go routine
+// allowing you to handle each item concurrently.  A maximum of limit walkFns will
+// be called at any one time.
+func WalkLimit(root string, options *godirwalk.Options, limit int) error {
+	// make sure limit is sensible
+	if limit < 1 {
+		panic("powerwalk: limit must be greater than zero.")
```
giuseppe: this should probably be an error, not a panic

go-selinux/pwalk.go
```diff
@@ -0,0 +1,128 @@
...
+			}
+		}
+		_ = godirwalk.Walk(root, &opts)
+
+		// everything is done
+		walkerWg.Done()
+	}()
+
+	// wait for all walker calls
+	walkerWg.Wait()
```
giuseppe: am I missing something or is there only one go routine to wait for?

go-selinux/pwalk.go
```diff
@@ -0,0 +1,128 @@
...
+	"sync"
+
+	"github.com/karrick/godirwalk"
+	"github.com/pkg/errors"
+)
+
+// DefaultConcurrentWalks is the default number of files that will be walked at the
+// same time when the Walk function is called.
+// To use a value other than this one, use the WalkLimit function.
+const DefaultConcurrentWalks int = 100
```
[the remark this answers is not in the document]
kolyshkin: A tool can make its own decision by using `WalkLimit()` function, which has this as a parameter. This is just a default, which I questioned as well but found sane enough.

I played a bit with various values for concurrency (on my laptop -- 8x cores, SSD), and they don't affect the performance much as long as the value is not less than the number of CPUs.
So, `100` looks like a sane default.

I'm not sure if it would be drastically different for rotating disks, as _I guess_ most if not all of the info required is already in the kernel's dentry cache. This would be interesting to play with though.

It might make sense to make this equal to e.g. `runtime.NumCPU() * 2`, iow

```go
func Walk(root string, options *godirwalk.Options) error {
￼	return WalkLimit(root, options, runtime.NumCPU() * 2)
￼}
```

go-selinux/pwalk.go
kolyshkin: agree

go-selinux/pwalk.go
kolyshkin: It is (and so a channel can be used instead).

I _guess_ the author played with having multiple walkers but decided to not go there, and this is left as a rudiment from that.

Will switch to using a channel.

kolyshkin:
Overall it seems it makes sense to fork the parallel walk package rather than sticking it in here.

go-selinux/pwalk.go
kolyshkin: Oh well, this can actually be simplified even further, i.e. there's no need for a goroutine

kolyshkin: In fact, let me simplify the whole thing ...

kolyshkin:
OK I rewrote the pwalk.go code to be much simpler and (I hope) comprehensible, also a bit faster.

@giuseppe @rhatdan @mrunalp PTAL

I plan to separate this code out to a small repo and use it here as well as in `containers/storage`.

giuseppe:
LGTM

go-selinux/pwalk.go
```diff
@@ -0,0 +1,95 @@
...
+// Walk walks the file tree rooted at root, calling options.Callback
+// for each file or directory in the tree, including root. All errors
+// that arise visiting files and directories are filtered by Callback.
+// The output is non-deterministic.
+//
+// For each file and directory encountered, Walk will trigger a new Go routine
+// allowing you to handle each item concurrently. A maximum of twice the
+// runtime.NumCPU() Callback goroutines will be called at any one time.
+func Walk(root string, options *godirwalk.Options) error {
+	return WalkLimit(root, options, runtime.NumCPU()*2)
```
rhatdan: I still think the number of threads should be customizable.

rhatdan:
If you want to move the walk code to a separate package, should we just put it in containers/conmmon?

go-selinux/pwalk.go
kolyshkin: It is (if you choose to use `WalkLimit()` function)

kolyshkin:
> should we just put it in containers/common?

That's the ideal place, thanks! Will open a PR once ready.

Closing this one for now.

PR #89, opened 2020-03-10
Add pkg/pwalk, make Chcon() faster

kolyshkin:
_This is a continuation of https://github.com/opencontainers/selinux/pull/85 and https://github.com/containers/common/pull/82_

**TL;DR:** This adds a parallel wrapper on top of `filetree.Walk()`, enabling multiple callbacks to be used concurrently. This is proven to speed things up. The resulting implementation is used for `Chcon()`, making it about 1.5x faster.

Please see more in commit messages. For previous discussions, see the abovementioned PRs.

NOTE that this is not as fast as https://github.com/opencontainers/selinux/pull/85. The reason is we use standard `filepath.Walk()` here, which performs the (unneeded) `os.Stat()` call on every entry, and also sorts the directory entries (which also does not make sense in many cases). So, yes, we can do better, but that means either another third-party package (such as github.com/karrick/godirwalk) or rewriting the filepath.Walk() from scratch. I'm still on the verge as whether we should do it, and am open to any suggestions.

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/opencontainers/selinux/pkg/pwalk"
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM 
But Lint is not happy.
@giuseppe @vrothberg @mrunalp PTAL

kolyshkin:
Once a release is made, I can pick this up in `containers/storage`

PR #93, opened 2020-04-04
Make sure selinux and nonselinux exported functions match

rhatdan:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,10 @@ define go-build
 	GOOS=$(1) GOARCH=$(2) $(GO) build -tags $(BUILDTAGS) ./...
 endef
 
+define go-build-noselinux
+	GOOS=$(1) GOARCH=$(2) $(GO) build ./...
+endef
+
 .PHONY:
 build:
 	$(call go-build,linux,amd64)
@@ -21,6 +25,13 @@ build-cross:
 	$(call go-build,linux,s390x)
 	$(call go-build,windows,amd64)
 	$(call go-build,windows,386)
+	$(call go-build-noselinux,linux,amd64)
+	$(call go-build-noselinux,linux,arm)
+	$(call go-build-noselinux,linux,arm64)
+	$(call go-build-noselinux,linux,ppc64le)
+	$(call go-build-noselinux,linux,s390x)
+	$(call go-build-noselinux,windows,amd64)
+	$(call go-build-noselinux,windows,386)
 
 BUILD_PATH := $(shell pwd)/build
 BUILD_BIN_PATH := ${BUILD_PATH}/bin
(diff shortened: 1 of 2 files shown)


rhatdan:
@giuseppe @vrothberg @kolyshkin @mrunalp @saschagrunert @thaJeztah PTAL

Makefile
```diff
@@ -32,7 +43,13 @@ ifndef GOPATH
 endif
 
 .PHONY: test
+noselinux:=$(shell mktemp /tmp/noselinux_XXXX)
+selinux:=$(shell mktemp /tmp/selinux_XXXX)
```
kolyshkin, line 47: These two files are created every time you run make...

Makefile
```diff
@@ -32,7 +43,13 @@ ifndef GOPATH
...
 test: check-gopath
+	grep '^func [A-Z].*' go-selinux/selinux_stub.go > ${noselinux}
+	grep '^func [A-Z].*' go-selinux/selinux_linux.go > ${selinux}
+	diff ${noselinux} ${selinux}
+	rm -f ${noselinux} ${selinux}
```
kolyshkin, line 52: but are only removed when you run `make test`

go-selinux/selinux_stub.go
```diff
@@ -113,7 +113,7 @@ SetTaskLabel sets the SELinux label for the current thread, or an error.
 This requires the dyntransition permission.
 */
 func SetTaskLabel(label string) error {
-        return nil
+	return nil
```
kolyshkin, line 116: NB: this is result of gofmt, replacing 8 spaces with a tab

Makefile
```diff
@@ -21,6 +25,13 @@ build-cross:
 	$(call go-build,linux,s390x)
 	$(call go-build,windows,amd64)
 	$(call go-build,windows,386)
+	$(call go-build-noselinux,linux,amd64)
+	$(call go-build-noselinux,linux,arm)
+	$(call go-build-noselinux,linux,arm64)
+	$(call go-build-noselinux,linux,ppc64le)
+	$(call go-build-noselinux,linux,s390x)
+	$(call go-build-noselinux,windows,amd64)
+	$(call go-build-noselinux,windows,386)
```
kolyshkin, line 34: since we're using gnu make extensions here already, maybe it makes sense to put these arches into a separate variable and use `$(foreach ...)` to avoid repetition (at least for linux)?

kolyshkin: e.g.
```make
linux-cross-arches := adm64 386 ....
$(foreach a, $(linux-cross-arches), $(eval $(call go-build, linux, $(a))))
```

rhatdan:
@kolyshkin Can you open a separate PR to fix the issues you are seeing. or I will.  But I want to get this in for fixes to Podman to continue.

PR #94, opened 2020-04-06
Implement FormatMountLabel unconditionally

thaJeztah:
Implementing FormatMountLabel on situations built without selinux should be possible; the `context` will be ignored if no SELinux is available.

### DIFF
--- a/go-selinux/label/label.go
+++ b/go-selinux/label/label.go
@@ -1,6 +1,8 @@
 package label
 
 import (
+	"fmt"
+
 	"github.com/opencontainers/selinux/go-selinux"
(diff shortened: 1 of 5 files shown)


thaJeztah:
@kolyshkin @rhatdan PTAL; this came up in a discussion recently on the containerd repository; in that discussion we considered this change (basically delegating this to runc and/or the platform instead of in the containerd source-code)

rhatdan:
LGTM

rhatdan:
@kolyshkin @mrunalp @giuseppe @vrothberg PTAL

PR #96, opened 2020-04-06
Fix Makefile to now build temp files when not temp

rhatdan:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,13 +43,8 @@ ifndef GOPATH
 endif
 
 .PHONY: test
-noselinux:=$(shell mktemp /tmp/noselinux_XXXX)
-selinux:=$(shell mktemp /tmp/selinux_XXXX)
 test: check-gopath
-	grep '^func [A-Z].*' go-selinux/selinux_stub.go > ${noselinux}
-	grep '^func [A-Z].*' go-selinux/selinux_linux.go > ${selinux}
-	diff ${noselinux} ${selinux}
-	rm -f ${noselinux} ${selinux}
+	bash -c "diff  <(grep '^func [A-Z]' go-selinux/selinux_stub.go) <(grep '^func [A-Z]' go-selinux/selinux_linux.go)"
 	go test -timeout 3m -tags "${BUILDTAGS}" ${TESTFLAGS} -v ./...
 	go test -timeout 3m ${TESTFLAGS} -v ./...
 


rhatdan:
@kolyshkin I could not get the foreach to work, in the limited time I played with it.  But this should fix the leaked tmpfiles.

Makefile
```diff
@@ -43,13 +43,13 @@ ifndef GOPATH
...
-noselinux:=$(shell mktemp /tmp/noselinux_XXXX)
-selinux:=$(shell mktemp /tmp/selinux_XXXX)
 test: check-gopath
-	grep '^func [A-Z].*' go-selinux/selinux_stub.go > ${noselinux}
-	grep '^func [A-Z].*' go-selinux/selinux_linux.go > ${selinux}
-	diff ${noselinux} ${selinux}
-	rm -f ${noselinux} ${selinux}
+	$(eval NOSELINUX := $(shell mktemp /tmp/noselinux_XXXX))
+	$(eval SELINUX := $(shell mktemp /tmp/selinux_XXXX))
+	grep '^func [A-Z].*' go-selinux/selinux_stub.go > ${NOSELINUX};
```
kolyshkin, line 49: It seems that `;` at EOL is redundant here (and in the lines below).

The alternative to is something like
```
diff <(grep '^func [A-Z].*' go-selinux/selinux_stub.go) <(grep ^func [A-Z].*' go-selinux/selinux_linux.go)
```
but it is bash/zsh specific

rhatdan: That is a nice elegant solution.

rhatdan: But getting it working in Makefile, will be difficult.

kolyshkin: why?
```make
checkme:
        bash -c "diff  <(grep '^func [A-Z].*' go-selinux/selinux_stub.go) <(grep '^func [A-Z].*' go-selinux/selinux_linux.go)"
```

```console
$ make  checkme
[kir@kir-rhat selinux]$ make checkme 
bash -c "diff  <(grep '^func [A-Z].*' go-selinux/selinux_stub.go) <(grep '^func [A-Z].*' go-selinux/selinux_linux.go)"
13a14
> func SetTaskLabel(label string) error {
make: *** [Makefile:35: checkme] Error 1
```

(i have deliberately removed `SetTaskLabel` for demo purposes)

kolyshkin: Just noticed that `.*` in grep regex is redundant

rhatdan: Because I am an idiot and was trying to do this in diff command rather then launching bash.

rhatdan:
I don't know where that comes from.

kolyshkin:
> ^ looks like pull approve needs updating; is that a configuration file somewhere?

I think I had it updated (there's that sync button on pullapprove dot com site) and it started to show my name in "needs 2 approval from ..." message, but apparently there's some more sorcery involved. @caniszczyk maybe you can help?

PR #98, opened 2020-05-21
Remote potential race condition, where mcs label is freed

rhatdan:
Theorectially if you do not change the MCS Label then we free it and two
commands later reserve it.  If some other process was grabbing MCS Labels
at the same time, the other process could get the same label.

Now we only release the label, if we have changed the MCSlevel.

Coverity was complaining about the selinux.GetEnabled() not being checked.


thaJeztah:
@kolyshkin ptal (I'm afk ☺️)

rhatdan:
@mrunalp @vrothberg PTAL

PR #100, opened 2020-06-05
Allow the category range to be changed

crosbymichael:
This keeps the range as part of the global state of the package but allows users
to modify the upper bounds of the category range.

Signed-off-by: Michael Crosby <michael@thepasture.io>

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -31,6 +31,9 @@ const (
 	// Disabled constant to indicate SELinux is disabled
 	Disabled = -1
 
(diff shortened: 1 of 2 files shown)


crosbymichael:
Travis passed, I'm not sure why GitHub hasn't updated the status

rhatdan:
Are you running on a system that does not have 1024 categories?

LGTM

crosbymichael:
@rhatdan I do have 1024 but want to limit the upper range for some reservations.

rhatdan:
OK
@mrunalp @saschagrunert @giuseppe @thaJeztah PTAL

PR #103, opened 2020-06-11
Add glblub implementation

yulicrunchy:
Policy version 32 (https://github.com/torvalds/linux/commit/42345b68c2e3e2b6549fc34b937ff44240dfc3b6) added support for greatest lower bound (glb) and least
upper bound (lub) calculations for a source and target range. This is a
port to the Go selinux library to provide support for users using a
kernel without glblub support.

Some example glblub calculations:

User Permitted Range | Network Device Label | Computed Label
---------------------|----------------------|----------------
s0-s1:c0.c12         | s0                   | s0
s0-s1:c0.c12         | s0-s1:c0.c1023       | s0-s1:c0.c12
s0-s4:c0.c512        | s1-s1:c0.c1023       | s1-s1:c0.c512
s0-s15:c0,c2         | s4-s6:c0.c128        | s4-s6:c0,c2
s0-s4                | s2-s6                | s2-s4
s0-s4                | s5-s8                | INVALID
s5-s8                | s0-s4                | INVALID

Signed-off-by: Yuli Khodorkovskiy <yuli@crunchydata.com>


go-selinux/selinux_linux.go
```diff
@@ -445,6 +460,186 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
 
+/*
+catsToBitset stores categories in a bitset.
+*/
+func catsToBitset(cats string) (*bitset.BitSet, error) {
+	bitset := &bitset.BitSet{}
+
+	catlist := strings.Split(cats, ",")
+	for _, r := range catlist {
+		ranges := strings.Split(r, ".")
```
rhatdan, line 471: This should use SplitN(r, ".", 2)

go-selinux/selinux_linux.go
```diff
@@ -445,6 +460,186 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+		if len(ranges) > 1 {
+			catstart, err := strconv.Atoi(string([]rune(ranges[0])[1:]))
```
rhatdan, line 473: I think you should use  strconv.ParseUint(, 10, 64) In order to catch bogus numbers.

go-selinux/selinux_linux.go
```diff
@@ -445,6 +460,186 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+
+	return bitset, nil
+}
+
+/*
+rangeStrToMLSRange marshals a string representation of a range.
+*/
+func rangeStrToMLSRange(rangeStr string) (*mlsRange, error) {
+	mlsRange := &mlsRange{}
+	levelSlice := strings.Split(rangeStr, "-")
```
rhatdan, line 501: Use SplitN(rangeStr,"-", 2)

go-selinux/selinux_linux.go
```diff
@@ -445,6 +460,186 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+
+	switch len(levelSlice) {
+	case 2:
+		high := strings.Split(levelSlice[1], ":")
+		sensHigh, err := strconv.Atoi(string([]rune(high[0])[1:]))
```
rhatdan, line 506: strconf.ParseUint()

go-selinux/selinux_linux_test.go
```diff
@@ -293,6 +293,91 @@ func TestComputeCreateContext(t *testing.T) {
...
 
+func TestGlbLub(t *testing.T) {
+	tests := []struct {
+		sourceRange   string
+		targetRange   string
+		expectedRange string
+		expectedErr   error
+	}{
+		{
+			sourceRange:   "s0:c0.c100-s10:c0.c150",
```
rhatdan, line 304: How about some bogus values.

go-selinux/selinux_linux.go
```diff
@@ -445,6 +460,186 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+func catsToBitset(cats string) (*bitset.BitSet, error) {
+	bitset := &bitset.BitSet{}
+
+	catlist := strings.Split(cats, ",")
+	for _, r := range catlist {
+		ranges := strings.Split(r, ".")
+		if len(ranges) > 1 {
+			catstart, err := strconv.Atoi(string([]rune(ranges[0])[1:]))
+			if err != nil {
+				return nil, err
```
rhatdan, line 475: Should some of these be errors.Wrap?  What do the errors look like when you give bogus data?

rhatdan: Should we check the catstart and catend to see if they are within the valid range of categories supported by the library? (1024)

rhatdan:
@saschagrunert @mrunalp @crosbymichael @kolyshkin @thaJeztah PTAL

go-selinux/selinux_linux.go
```diff
@@ -445,6 +460,186 @@ func ComputeCreateContext(source string, target string, class string) (string, e
 	return readWriteCon(filepath.Join(getSelinuxMountPoint(), "create"), fmt.Sprintf("%s %s %d", source, target, classidx))
 }
 
+/*
+catsToBitset stores categories in a bitset.
+*/
```
saschagrunert, line 465:
```suggestion
// catsToBitset stores categories in a bitset.
```

go-selinux/selinux_linux.go
```diff
@@ -445,6 +460,186 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+			bitset.Set(uint(cat))
+		}
+	}
+
+	return bitset, nil
+}
+
+/*
+rangeStrToMLSRange marshals a string representation of a range.
+*/
```
saschagrunert, line 498:
```suggestion
// rangeStrToMLSRange marshals a string representation of a range.
```

go-selinux/selinux_linux.go
yulicrunchy: > Should some of these be errors.Wrap? What do the errors look like when you give bogus data?

Thanks for the review.

Any objection to using `fmt.Errorf()` and `errors.Is()` for wrapping? 



>Should we check the catstart and catend to see if they are within the valid range of categories supported by the library? (1024)

We have use cases where we need more than 1024 categories. Would you prefer a maxCats  argument to this function or to parse this from `/sys/fs/selinux/initial_contexts/kernel`?

go.mod
```diff
@@ -4,5 +4,6 @@ go 1.13
 
 require (
 	github.com/pkg/errors v0.8.1
+	github.com/willf/bitset v1.1.10
```
thaJeztah, line 7: This dependency is BSD-3 Licensed; I _think_ that's ok, but thought I'd add a comment for visibility https://github.com/willf/bitset

go-selinux/selinux_linux.go
rhatdan: I would prefer to continue using errors.Wrapf, since the code already uses it.

kolyshkin: I think eventually we should switch to `fmt.Errorf` and `errors.Is/.As` since it's now a standard which does not require a third-party package. Having said that, such a change needs to be performed at once and repo-wise.

go-selinux/selinux_linux.go
```diff
@@ -445,6 +460,186 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+	return fmt.Sprintf("%s-%s", low, high)
+}
+
+/*
+CalculateGlbLub computes the glb (greatest lower bound) and lub (least upper bound)
+of a source and target range.
+The glblub is calculated as the greater of the low sensitivities and
+the lower of the high sensitivities and the and of each category bitset.
+*/
+func CalculateGlbLub(sourceRange string, targetRange string) (string, error) {
```
kolyshkin, line 609: nit: `sourceRange, targetRange string`

go-selinux/selinux_linux.go
```diff
@@ -445,6 +460,186 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+	if reflect.DeepEqual(m.low, m.high) {
+		return low
+	}
+
+	high := fmt.Sprintf("s%d", m.high.sens)
+	if m.high.cats != nil && m.high.cats.Count() > 0 {
+		high = fmt.Sprintf("%s:%s", high, bitsetToStr(m.high.cats))
+	}
+
+	return fmt.Sprintf("%s-%s", low, high)
```
kolyshkin, line 600: nit: this one (and a few `Sprintf`s above can be mere string concatenations, e.g. `low + "-" + high`. Printf is slow in Go.

yulicrunchy:
I pushed the requested updates. Please lmk if another update is required.

rhatdan:
LGTM
@kolyshkin PTAL

go-selinux/selinux_linux.go
```diff
@@ -445,6 +469,194 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+	item := []rune(s)
+	if item[0] != rune(sep) {
+		return 0, ErrLevelSyntax
+	}
+	val, err := strconv.ParseUint(string(item)[1:], 10, 64)
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to parse level")
+	}
+
+	return uint32(val), nil
```
kolyshkin, line 514: Can you elaborate on why you parse the number as uint64, then convert it to uint32, then convert it back to uint[64] (when calling `bitset.Set()`)? If these are supposed to be 32-bit numbers, we should use `strconv.ParseUint(..., 32)` for conversion to catch potential overflows.

kolyshkin:
I am slightly opposed to using `rune` and conversion of `string` to `[]rune` in here. Since the input strings should only contains ASCII, we don't have to aim for UTF-8 compatibility here, and it's faster to use string/byte.

I wrote a couple of quick tests, mimicking the `parseLevelItem`.

1. Using rune:

```go
package main

import (
	"strconv"
	"testing"
)

type levelItem rune

const (
	sensitivity levelItem = 's'
	category    levelItem = 'c'
)

func check(s string, sep levelItem) bool {
	item := []rune(s)
	if item[0] != rune(sep) {
		return false
	}
	_, err := strconv.ParseUint(string(item)[1:], 10, 64)
	return err == nil
}

func BenchmarkRune(b *testing.B) {
	x := "s2345"
	var res bool
	for i := 0; i < b.N; i++ {
		res = check(x, sensitivity)
	}
	if !res {
		b.Fail()
	}
}
```

result:
```console
$ go test -v -bench .
goos: linux
goarch: amd64
pkg: github.com/opencontainers/runc/tst
BenchmarkRune
BenchmarkRune-8   	16515205	        69.8 ns/op
PASS
ok  	github.com/opencontainers/runc/tst	1.229s
```

2. Using string/byte:

```go
package main

import (
	"strconv"
	"testing"
)

type levelItem byte

const (
	sensitivity levelItem = 's'
	category    levelItem = 'c'
)

func check(s string, sep levelItem) bool {
	if levelItem(s[0]) != sep {
		return false
	}
	_, err := strconv.ParseUint(s[1:], 10, 64)
	return err == nil
}

func BenchmarkString(b *testing.B) {
	x := "s2345"
	var res bool
	for i := 0; i < b.N; i++ {
		res = check(x, sensitivity)
	}
	if !res {
		b.Fail()
	}
}
```

Result:
```console
$ go test -v -bench .
goos: linux
goarch: amd64
pkg: github.com/opencontainers/runc/tst/x
BenchmarkString
BenchmarkString-8   	73324179	        16.5 ns/op
PASS
ok  	github.com/opencontainers/runc/tst/x	1.229s
```

As you can see, string/byte is about 4x faster.

go-selinux/selinux_linux.go
```diff
@@ -445,6 +469,194 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+
+	if s.high.sens < t.low.sens || t.high.sens < s.low.sens {
+		/* these ranges have no common sensitivities */
+		return "", ErrIncomparable
+	}
+
+	outrange := &mlsRange{low: &level{}, high: &level{}}
+
+	/* take the greatest of the low */
+	outrange.low.sens = uint32(math.Max(float64(s.low.sens), float64(t.low.sens)))
```
kolyshkin, line 644: The `int` to `float` (and back) conversion just to use `Max()` looks unholy to me. It's probably very slow and prone to errors.

Until we have generics or `?:` operator in Go, I think in this case it would be the best to write your own min and max functions.

go-selinux/selinux_linux.go
```diff
@@ -445,6 +469,194 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+}
+
+// String returns an mlsRange as a string.
+func (m mlsRange) String() string {
+	low := fmt.Sprintf("s%d", m.low.sens)
+	if m.low.cats != nil && m.low.cats.Count() > 0 {
+		low += ":" + bitsetToStr(m.low.cats)
+	}
+
+	if reflect.DeepEqual(m.low, m.high) {
```
kolyshkin, line 608: This (`reflect.DeepEqual`) is probably too costly to use anywhere expect for tests (or when you don't know the data format you're dealing with). Please write a comparison function.

go-selinux/selinux_linux.go
yulicrunchy: 👍 will do

go-selinux/selinux_linux.go
yulicrunchy: will do

go-selinux/selinux_linux.go
yulicrunchy: This was just an oversight after looking at the Go documentation for `ParseUint()`. Following your suggestion for a custom Max/Min, this should also get rid of some of the unnecessary conversions.

yulicrunchy:
> ```
> 
> As you can see, string/byte is about 4x faster.

@kolyshkin thanks for the review and suggestions. I'll get the changes pushed up as soon as I can.

yulicrunchy:
@kolyshkin I have made the requested changes with some additional fixes:

- switched from `bytes.Buffer` to the `+` string concatanation. I wasn't checking `buffer.WriteString()` for errors and this seemed cleaner
- updated the bitset package to point to master. There was a bug that was addressed when calling the `Equal()` operator on a nil BitSet

go-selinux/selinux_linux.go
```diff
@@ -445,6 +467,217 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+			bitset.Set(cat)
+		}
+	}
+
+	return bitset, nil
+}
+
+// parseLevelItem parses and verifies that a sensitivity or category are valid
+func parseLevelItem(s string, sep levelItem) (uint, error) {
+	if levelItem(s[0]) != sep {
```
kolyshkin, line 503: `s[0]` here can panic on an empty string, and `s[1:]` below can panic on a string with len(str) > 2, so maybe something like

```suggestion
	if len(s) < 2 || levelItem(s[0]) != sep {
```

go-selinux/selinux_linux.go
```diff
@@ -445,6 +467,217 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+		return 0, errors.Wrap(err, "failed to parse level")
+	}
+
+	return uint(val), nil
+}
+
+// rangeStrToMLSRange marshals a string representation of a range.
+func rangeStrToMLSRange(rangeStr string) (*mlsRange, error) {
+	mlsRange := &mlsRange{}
+	levelSlice := strings.SplitN(rangeStr, "-", 2)
```
kolyshkin, line 517: nit: adding a few examples of how `rangeStr` might look like would help in review and future development (surely it is available but you need to look elsewhere)

go-selinux/selinux_linux.go
```diff
@@ -445,6 +467,217 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+		mlsRange.high = mlsRange.low
+	}
+
+	return mlsRange, nil
+}
+
+/*
+bitsetToStr takes a category bitset and returns it in the
+canonical selinux syntax
+*/
```
kolyshkin, line 561: nit: maybe use `//`-style comment here. I'm not exactly sure why but `/* ... */` is not widely popular/used in Go

go-selinux/selinux_linux.go
```diff
@@ -445,6 +467,217 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+func bitsetToStr(c *bitset.BitSet) string {
+	var catBuff string
```
kolyshkin, line 563: nit: `s/catBuff/cat/` (or `str`, or `ret`).

go-selinux/selinux_linux.go
```diff
@@ -445,6 +467,217 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+		len = 0
+		i = next
+	}
+
+	return catBuff
+}
+
+func (l1 *level) equal(l2 *level) bool {
+	if l2 == nil || l1 == nil {
+		return l1 == l2
```
kolyshkin, line 600: nice :)

go-selinux/selinux_linux.go
```diff
@@ -445,6 +467,217 @@ func ComputeCreateContext(source string, target string, class string) (string, e
...
+	}
+	if l1.sens != l2.sens {
+		return false
+	}
+	return l1.cats.Equal(l2.cats)
+}
+
+// String returns an mlsRange as a string.
+func (m mlsRange) String() string {
+	low := fmt.Sprintf("s%d", m.low.sens)
```
kolyshkin, line 610: nit:
```suggestion
	low := "s" + strconv.Itoa(m.low.sens)
```

go-selinux/selinux_linux.go
```diff
@@ -52,13 +55,32 @@ type selinuxState struct {
...
+type level struct {
+	sens uint
+	cats *bitset.BitSet
+}
+
+type mlsRange struct {
+	low  *level
+	high *level
+}
+
```
kolyshkin, line 67: nit: maybe move `type levelItem` definition here

kolyshkin:
Looks way better now, thanks! Left a couple of minor nits, feel free to ignore any :)

go-selinux/selinux_linux.go
yulicrunchy: done

go-selinux/selinux_linux.go
yulicrunchy: done

go-selinux/selinux_linux.go
yulicrunchy: done

go-selinux/selinux_linux.go
yulicrunchy: done

go-selinux/selinux_linux.go
yulicrunchy: I added a few. Let me know if they are sufficient.

go-selinux/selinux_linux.go
yulicrunchy: Good catch. Fixed.

yulicrunchy:
@kolyshkin I pushed another iteration. Beyond the fixes you suggested I moved some common code into `parseLevel()` and added more context to parsing errors. Thanks for the review.

thaJeztah:
@yulicrunchy I haven't looked at the latest changes yet, but can you squash the commits? (I suspect having the `vendor` commit separate will break git bisect)

yulicrunchy:
> @yulicrunchy I haven't looked at the latest changes yet, but can you squash the commits? (I suspect having the `vendor` commit separate will break git bisect)

done

go-selinux/selinux_linux_test.go
```diff
@@ -293,6 +296,126 @@ func TestComputeCreateContext(t *testing.T) {
...
+			expectedErr: ErrLevelSyntax,
+		},
+		{
+			sourceRange: "s5:x50.x100-s15:c0.c149",
+			targetRange: "s5:c50.c100-s15:c0.c149",
+			expectedErr: ErrLevelSyntax,
+		},
+	}
+
+	if !GetEnabled() {
```
kolyshkin, line 393: Not sure you need this here, as you don't actually use any of selinux functionality in this test. Probably a copy-paste leftover.

kolyshkin:
A single nit about the test, otherwise LGTM

go-selinux/selinux_linux_test.go
yulicrunchy: Removed and pushed. Thanks.

kolyshkin:
LGTM

thaJeztah:
I noticed this PR switched from a tagged release of willf/bitset to master; https://github.com/opencontainers/selinux/pull/103#issuecomment-651861843

> pinned the bitset package version to point to master as of this morning. There was a bug that was addressed when calling the Equal() operator on a nil BitSet

I opened https://github.com/willf/bitset/issues/83 to ask if a new release can be tagged

PR #113, opened 2020-07-22
Clarify documentation for Set{Task,Exec}Label

yulicrunchy:
In Go, there are no guarantees which goroutine runs in which thread.
It is therefore important to wrap calls to Set{Task,Exec}Label with
runtime.LockOSThread() and runtime.UnlockOSThread() to ensure reliable
behavior of goroutine labeling.

### DIFF
--- a/go-selinux/selinux.go
+++ b/go-selinux/selinux.go
@@ -63,8 +63,12 @@ func FileLabel(fpath string) (string, error) {
 	return fileLabel(fpath)
 }
 
-// SetFSCreateLabel tells kernel the label to create all file system objects
-// created by this task. Setting label="" to return to default.
+// SetFSCreateLabel tells the kernel what label to use for all file system objects
+// created by this task.
+// Set the label to an empty string to return to the default label. Calls to SetFSCreateLabel
+// should be wrapped in runtime.LockOSThread()/runtime.UnlockOSThread() until file system
+// objects created by this task are finished to guarantee another goroutine does not migrate
+// to the current thread before execution is complete.
 func SetFSCreateLabel(label string) error {
 	return setFSCreateLabel(label)
 }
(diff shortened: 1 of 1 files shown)


rhatdan:
There are more calls then just these that need your comment.

Also you need to sign your PR.
git commit -a --amend -s
git push --force

rhatdan:
SetFSCreateLabel(label string)
SetExecLabel
SetTaskLabel
SetSocketLabel
SetKeyLabel

rhatdan:
LGTM
@kolyshkin @mrunalp @thaJeztah @crosbymichael PTAL

go-selinux/selinux.go
```diff
@@ -63,8 +63,11 @@ func FileLabel(fpath string) (string, error) {
 	return fileLabel(fpath)
 }
 
-// SetFSCreateLabel tells kernel the label to create all file system objects
-// created by this task. Setting label="" to return to default.
+// SetFSCreateLabel tells the kernel the label to create all file system objects
+// created by this task. Set label="" to return to the default label. Calls to SetFSCreateLabel
+// should be wrapped in runtime.LockOSThread()/runtime.UnlockOSThread() until file system
+// objects created by this task are finished to guarantee another goroutine does not migrate
+// to the current thread before execution is complete.
```
thaJeztah, line 70: Not a blocker (as it was already a bit confusing), but the first line is a bit hard to grasp. Perhaps something like;

```suggestion
// SetFSCreateLabel tells the kernel what label to use for all file system objects created by this task.
// Set label to an empty string to return to the default label. Calls to SetFSCreateLabel
// should be wrapped in runtime.LockOSThread()/runtime.UnlockOSThread() until file system
// objects created by this task are finished to guarantee another goroutine does not migrate
// to the current thread before execution is complete.
```

(suggestions welcome! /cc @rhatdan)

rhatdan:
@yulicrunchy Please take @thaJeztah input and then we can merge.

thaJeztah:
LGTM

thaJeztah:
@rhatdan @kolyshkin PTAL

rhatdan:
Thanks @yulicrunchy.  This is an issue that often trips people up, including me.

PR #115, opened 2020-08-12
travis: add go 1.15, rm unsupported versions

kolyshkin:
Since go 1.15 is out, go 1.13 is no longer supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 go:
-  - 1.12.x
-  - 1.13.x
+  - 1.14.x
+  - 1.15.x
   - tip
 
 before_install:


rhatdan:
LGTM
@mrunalp @saschagrunert @thaJeztah @vrothberg PTAL

thaJeztah:
Would it be possible to keep 1.13 for a short while? Go 1.14 has proven to be a problematic release, and while downstream projects will upgrade to Go 1.15 (which fixes most of the issues with Go 1.14), not all project have migrated yet (and may wait for the first patch release of Go 1.15 in case there's regressions).

> change the go.mod version too.

IIRC, setting a higher go version in go.mod will result in consumers of the package to no longer being able to build, correct? If so, I wonder if we should be somewhat conservative there as well, and not _restrict_ older versions (unless we have specific reasons to not allow other versions).

saschagrunert:
> > change the go.mod version too.
> 
> IIRC, setting a higher go version in go.mod will result in consumers of the package to no longer being able to build, correct? If so, I wonder if we should be somewhat conservative there as well, and not _restrict_ older versions (unless we have specific reasons to not allow other versions).

I think the version is more an indicator to the module system if there are some changes needed to the vendored sources or go.mod/sum. It does not restrict the build as far as I know,.

thaJeztah:
> I think the version is more an indicator to the module system if there are some changes needed to the vendored sources or go.mod/sum. It does not restrict the build as far as I know,.

Ah, looks like that may have been a bug in Go < 1.11.4 https://github.com/golang/go/issues/30446#issuecomment-468029913 (see https://github.com/golang/go/commit/e32203f647370897c6a28018c16cfd9584849569)

I wonder if they start enforcing at some point though (same as they started enforcing the import paths for Go 1.13 and up 🤔)

rhatdan:
Well if this becomes a problem we can always revert.

PR #117, opened 2020-09-17
make vendor: fix

kolyshkin:
This fixes Makefile's `vendor` target, and updates the vendor.

Please see individual commits for details. Spoiler alert: some commits are rather entertaining.


vrothberg (commented):
LGTM

`GO111MODULE=off` was set to make sure that all Go commands use the vendor directory instead of fetching the dependencies to the module cache. Since 1.14 Go defaults to `-mod=vendor` when there's a ./vendor directory which is a very welcome change.

rhatdan:
LGTM

kolyshkin:
> GO111MODULE=off was set to make sure that all Go commands use the vendor directory.

right, thanks!

PR #119, opened 2020-09-25
Implement get_default_context_with_level() from libselinux

yulicrunchy:
This PR:

- implements get_default_context_with_level() from libselinux.
- change parser code for the SELinux config and label files to use bufio.Scanner
- cache the SELinux root policy dir value

### DIFF
--- a/go-selinux/selinux.go
+++ b/go-selinux/selinux.go
@@ -30,6 +30,11 @@ var (
 	// ErrLevelSyntax is returned when a sensitivity or category do not have correct syntax in a level
 	ErrLevelSyntax = errors.New("invalid level syntax")
 
+	// ErrContextMissing is returned if a requested context is not found in a file.
+	ErrContextMissing = errors.New("context does not have a match")
+	// ErrVerifierNil is returned when a context verifier function is nil.
+	ErrVerifierNil = errors.New("verifier function is nil")
+
 	// CategoryRange allows the upper bound on the category range to be adjusted
 	CategoryRange = DefaultCategoryRange
 )
@@ -261,4 +266,13 @@ func DupSecOpt(src string) ([]string, error) {
 // labeling support for future container processes.
 func DisableSecOpt() []string {
 	return disableSecOpt()
+}
+
+// GetDefaultContextWithLevel gets a single context for the specified SELinux user
+// identity that is reachable from the specified scon context. The context is based
+// on the per-user /etc/selinux/{SELINUXTYPE}/contexts/users/<username> if it exists,
+// and falls back to the global /etc/selinux/{SELINUXTYPE}/contexts/default_contexts
+// file.
+func GetDefaultContextWithLevel(user, level, scon string) (string, error) {
+	return getDefaultContextWithLevel(user, level, scon)
 }
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -207,28 +207,16 @@ func getEnabled() bool {
 }
 
 func readConfig(target string) string {
-	var (
-		val, key string
-		bufin    *bufio.Reader
-	)
-
 	in, err := os.Open(selinuxConfig)
 	if err != nil {
 		return ""
 	}
 	defer in.Close()
 
-	bufin = bufio.NewReader(in)
+	scanner := bufio.NewScanner(in)
 
-	for done := false; !done; {
-		var line string
-		if line, err = bufin.ReadString('\n'); err != nil {
-			if err != io.EOF {
-				return ""
-			}
-			done = true
-		}
-		line = strings.TrimSpace(line)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
 		if len(line) == 0 {
 			// Skip blank lines
 			continue
@@ -238,7 +226,7 @@ func readConfig(target string) string {
 			continue
 		}
 		if groups := assignRegex.FindStringSubmatch(line); groups != nil {
-			key, val = strings.TrimSpace(groups[1]), strings.TrimSpace(groups[2])
+			key, val := strings.TrimSpace(groups[1]), strings.TrimSpace(groups[2])
 			if key == target {
 				return strings.Trim(val, "\"")
 			}
@@ -889,11 +877,6 @@ func openContextFile() (*os.File, error) {
 var labels = loadLabels()
 
 func loadLabels() map[string]string {
-	var (
-		val, key string
-		bufin    *bufio.Reader
-	)
-
 	labels := make(map[string]string)
 	in, err := openContextFile()
 	if err != nil {
@@ -901,18 +884,10 @@ func loadLabels() map[string]string {
 	}
 	defer in.Close()
 
-	bufin = bufio.NewReader(in)
+	scanner := bufio.NewScanner(in)
 
-	for done := false; !done; {
-		var line string
-		if line, err = bufin.ReadString('\n'); err != nil {
-			if err == io.EOF {
-				done = true
-			} else {
-				break
-			}
-		}
-		line = strings.TrimSpace(line)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
 		if len(line) == 0 {
 			// Skip blank lines
 			continue
@@ -922,7 +897,7 @@ func loadLabels() map[string]string {
 			continue
 		}
 		if groups := assignRegex.FindStringSubmatch(line); groups != nil {
-			key, val = strings.TrimSpace(groups[1]), strings.TrimSpace(groups[2])
+			key, val := strings.TrimSpace(groups[1]), strings.TrimSpace(groups[2])
 			labels[key] = strings.Trim(val, "\"")
 		}
 	}
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -28,6 +28,8 @@ const (
 	minSensLen       = 2
 	contextFile      = "/usr/share/containers/selinux/contexts"
 	selinuxDir       = "/etc/selinux/"
+	selinuxUsersDir  = "contexts/users"
+	defaultContexts  = "contexts/default_contexts"
 	selinuxConfig    = selinuxDir + "config"
 	selinuxfsMount   = "/sys/fs/selinux"
 	selinuxTypeTag   = "SELINUXTYPE"
@@ -54,6 +56,13 @@ type level struct {
 type mlsRange struct {
 	low  *level
 	high *level
+}
+
+type defaultSECtx struct {
+	user, level, scon   string
+	userRdr, defaultRdr io.Reader
+
+	verifier func(string) error
 }
 
 type levelItem byte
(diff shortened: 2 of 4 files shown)


rhatdan:
LGTM
@thaJeztah @kolyshkin @vrothberg PTAL

rhatdan:
@wrabcak PTAL

thaJeztah:
I started reviewing this, may have some comments, but will try to get back to this soon 😅 

/cc @kolyshkin

go-selinux/selinux.go
```diff
@@ -262,3 +262,10 @@ func DupSecOpt(src string) ([]string, error) {
 func DisableSecOpt() []string {
 	return disableSecOpt()
 }
+
+// GetDefaultContextWithLevel gets a single context for the specified SELinux user identity that is reachable from the specified scon context.
+// The contetext is based on  the  global  /etc/selinux/{SELINUXTYPE}/contexts/default_contexts file and the per-user
```
kolyshkin, line 267: `s/contetext/context/`

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1104,115 @@ func dupSecOpt(src string) ([]string, error) {
 func disableSecOpt() []string {
 	return []string{"disable"}
 }
+
+func getDefaultContextFromReaders(c defaultSECtx) (string, error) {
+	if c.verifier == nil {
+		return "", fmt.Errorf("verifier function must not be nil")
```
kolyshkin, line 1110: nit: use `errors.New` as there's nothing to format here.

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1104,115 @@ func dupSecOpt(src string) ([]string, error) {
...
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to create label for %s", c.scon)
+	}
+
+	findUserInContext := func(r io.Reader) string {
+		fromRole := context["role"]
+		fromType := context["type"]
+		bufin := bufio.NewReader(r)
+
+		for done := false; !done; {
```
kolyshkin, line 1123: I think it will be easier/cleaner to use bufio scanner here:
```go
s := bufio.NewScanner(r)
for scanner.Scan() {
        line = strings.TrimSpace(s.Text())
        ...
}
if err := scanner.Err(); err != nil {
        // FIXME log/return err?
        return ""
}

File: go-selinux/selinux_linux.go [hunk 4]
```diff
@@ -1095,3 +1104,115 @@ func dupSecOpt(src string) ([]string, error) {
...
+	if c.verifier == nil {
+		return "", fmt.Errorf("verifier function must not be nil")
+	}
+
+	context, err := newContext(c.scon)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to create label for %s", c.scon)
+	}
+
+	findUserInContext := func(r io.Reader) string {
```
kolyshkin, line 1118: Maybe separate it out as a function? Somewhat harder to read otherwise.

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1104,115 @@ func dupSecOpt(src string) ([]string, error) {
...
+			if len(line) == 0 {
+				// Skip blank lines
+				continue
+			}
+			if line[0] == ';' || line[0] == '#' {
+				// Skip comments
+				continue
+			}
+
+			conns := strings.Fields(line)
```
kolyshkin, line 1146: Since you have used TrimSpace already, it might make sense to use strings.Split here (it's 2x faster than strings.Fields).

If your input is more complicated (various delimiters, may have 2 spaces instead of one etc.), use `strings.Fields` but in this case you don't need `strings.TrimSpace` beforehand.

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1104,115 @@ func dupSecOpt(src string) ([]string, error) {
...
+
+			lineConn := strings.TrimSpace(conns[0])
+			lineArr := strings.SplitN(lineConn, ":", 3)
+
+			// malformed context in the context file
+			if len(lineArr) != 3 {
```
kolyshkin, line 1152: So, unless you expect `lineArr[2]` to have semicolons, you need to do
```go
lineArr := strings.SplitN(lineConn, ":", 4)
if len(lineArr) != 3 {
```

to catch the "too many semicolons" type of error

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1104,115 @@ func dupSecOpt(src string) ([]string, error) {
...
+				continue
+			}
+			if line[0] == ';' || line[0] == '#' {
+				// Skip comments
+				continue
+			}
+
+			conns := strings.Fields(line)
+
+			lineConn := strings.TrimSpace(conns[0])
```
kolyshkin, line 1148: Again, since you have used `strings.Fields` above, the `conns[0]` is guaranteed to not have extra spaces at the edges, so `strings.TrimSpace` here is obviously redundant.

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1104,115 @@ func dupSecOpt(src string) ([]string, error) {
...
+			if len(lineArr) != 3 {
+				continue
+			}
+
+			if lineArr[0] != fromRole || lineArr[1] != fromType {
+				continue
+			}
+
+			for _, cc := range conns[1:] {
+				userConn := strings.TrimSpace(cc)
```
kolyshkin, line 1161: same about `strings.TrimSpace`

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1104,115 @@ func dupSecOpt(src string) ([]string, error) {
...
+				userArr := strings.SplitN(userConn, ":", 3)
```
kolyshkin, line 1162: 1. This should be taken out of the `for` loop, otherwise you're doing the same thing again and again.

2. No check for a number of fields returned.

3. See https://github.com/opencontainers/selinux/pull/119/commits/6fd60bb08ac5d42cad52be3d3195e97464b57585#r497819835

go-selinux/selinux_linux.go
kolyshkin: BTW feel free to rewrite `readConfig` (from which you apparently copy-pasted this code) to use bufio.Scanner, too (in a separate commit).

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1104,115 @@ func dupSecOpt(src string) ([]string, error) {
...
+				context["role"] = userArr[0]
+				context["type"] = userArr[1]
+				context["level"] = c.level
+
+				outConn := context.get()
+				if err := c.verifier(outConn); err != nil {
+					continue
+				}
+
+				return outConn
```
kolyshkin, line 1174: Overall I think this should return errors, rather than silently ignore all input it can't process.

kolyshkin: Or some of it... I dunno

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1104,115 @@ func dupSecOpt(src string) ([]string, error) {
...
+	if conn := findUserInContext(c.defaultRdr); conn != "" {
+		return conn, nil
+	}
+
+	return "", fmt.Errorf("could not find context match for %q", c.scon)
+}
+
+func getDefaultContextWithLevel(user, level, scon string) (string, error) {
+	userPath := filepath.Join(getSELinuxPolicyRoot(), selinuxUsersDir, user)
+	defaultPath := filepath.Join(getSELinuxPolicyRoot(), defaultContexts)
```
kolyshkin, line 1194: Funny that every line here causes re-reading of `/etc/selinux/config`. At the very least you can reuse the `getSELinuxPolicyRoot` result here.

Feel free to optimize that in the existing code as well (separate commit).

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1104,115 @@ func dupSecOpt(src string) ([]string, error) {
 func disableSecOpt() []string {
 	return []string{"disable"}
 }
+
+func getDefaultContextFromReaders(c defaultSECtx) (string, error) {
```
kolyshkin, line 1108: I think you can use a pointer as an argument here, to save some stack space.

go-selinux/selinux_linux_test.go
```diff
@@ -412,6 +413,76 @@ func TestGlbLub(t *testing.T) {
...
+			t.Fatalf("err should not exist but is: %v", err)
+		}
+
+		if got != want {
+			t.Fatalf("got context: %q but expected %q", got, want)
+		}
+	})
+
+	t.Run("match exists in default context file, but not in user file", func(t *testing.T) {
+		userBuff = `# COMMENT
```
kolyshkin, line 458: You rewrite the `userBuff` here, so all the test cases below will use it. This smells buggy to me.

go-selinux/selinux_linux_test.go
```diff
@@ -412,6 +413,76 @@ func TestGlbLub(t *testing.T) {
...
+foo_r:foo_t:s0     sysadm_r:sysadm_t:s0
+staff_r:staff_t:s0                 baz_r:baz_t:s0   sysadm_r:sysadm_t:s0
+`
+
+	c := defaultSECtx{
+		user:       "bob",
+		level:      "SystemLow-SystemHigh",
+		scon:       "system_u:staff_r:staff_t:s0",
+		userRdr:    bytes.NewBufferString(userBuff),
+		defaultRdr: bytes.NewBufferString(defaultBuff),
```
kolyshkin, line 442: So, these are getting consumed by every test, and I don't see any resets here...

kolyshkin (changes requested):
left plenty of comments

wrabcak:
> @wrabcak PTAL

Reassigning to @bachradsusi.

bachradsusi:
I don't speak GO, but what I see seems to follow what libselinux does - generate a context from user and contexts matching scon from contexts/users and check whether the generated context is valid. It's important it doesn't use /sys/fs/selinux/user interface directly or via deprecated security_compute_user()

I'd only appreciate if the commit message contain more details but it's up to the project habits.

go-selinux/selinux_linux.go
yulicrunchy: > This should be taken out of the for loop, otherwise you're doing the same thing again and again

Are you referring to the setting of user and level in the context or something else?

> See 6fd60bb#r497819835

This URL doesn't work for some reason. I'm assuming you meant setting the subslices in `SplitN` to 4?

go-selinux/selinux_linux.go
yulicrunchy: In the libselinux implementation, they loop until a valid match is found, silently ignoring the invalid contexts. Maybe it is the right thing to do to return an error if the entry is invalid, since the user should be aware of invalid contexts, but we lose feature parity with libselinux. I defer to you on what you want me to do.

go-selinux/selinux_linux.go
yulicrunchy: >  it might make sense to use strings.Split here (it's 2x faster than strings.Fields).
Is there recommended reading you can point me to about performant Go? Or is this something you just know from experience? Thanks

go-selinux/selinux_linux.go
bachradsusi: The idea in libselinux is that if there's a typo in one context, it is better to skip the wrong entry and try to parse the rest.

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1077,122 @@ func dupSecOpt(src string) ([]string, error) {
...
+		lineArr := strings.SplitN(line, ":", 4)
+		// skip context with typo, or role and type do not match
+		if len(lineArr) != 3 ||
+			lineArr[0] != fromRole ||
+			lineArr[1] != fromType {
+			continue
+		}
+
+		for _, cc := range fromConns[1:] {
+			toConns := strings.SplitN(cc, ":", 4)
```
rhatdan, line 1113: Why not use newContext here rather then reimplementing it?

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1077,122 @@ func dupSecOpt(src string) ([]string, error) {
...
+		line := fromConns[0]
+
+		if line[0] == ';' || line[0] == '#' {
+			// Skip comments
+			continue
+		}
+
+		lineArr := strings.SplitN(line, ":", 4)
+		// skip context with typo, or role and type do not match
+		if len(lineArr) != 3 ||
```
rhatdan, line 1106: Why not use newContext here rather then reimplementing it?  Not sure what the lineArr !3 is for?

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1077,122 @@ func dupSecOpt(src string) ([]string, error) {
...
+	context["user"] = c.user
+	context["level"] = c.level
+
+	conn, err := findUserInContext(context, c.userRdr, c.verifier)
+	if err != nil {
+		return "", err
+	}
+
+	if conn != "" {
+		return conn, err
```
rhatdan, line 1157: Return nil to make it clear to future developers that this is success.

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1077,122 @@ func dupSecOpt(src string) ([]string, error) {
...
+	}
+
+	conn, err = findUserInContext(context, c.defaultRdr, c.verifier)
+	if err != nil {
+		return "", err
+	}
+
+	if conn != "" {
+		return conn, err
```
rhatdan, line 1166: Return nil to make it clear to future developers that this is success.

go-selinux/selinux_linux.go
```diff
@@ -1095,3 +1077,122 @@ func dupSecOpt(src string) ([]string, error) {
...
+	}
+
+	return "", errors.Errorf("could not find context match for: %q", c.scon)
```
rhatdan, line 1169: We really should start making some constants.  Not something this PR needs to change, but it would be nice for callers into SELinux to compare errors to error codes.

rhatdan (approved):

go-selinux/selinux_linux.go
yulicrunchy: I'll declare constants for new errors introduced in this PR.

rhatdan (approved):
LGTM

rhatdan:
@kolyshkin @thaJeztah @vrothberg PTAL

rhatdan:
Second attempt
@kolyshkin @thaJeztah @vrothberg PTAL

rhatdan:
I would like to cut a release and get this in.

vrothberg (commented):
LGTM

PR #120, opened 2020-10-14
 Add poweron architecture ppc64le to travis build

asellappen:
This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing, For more info please tag @gerrith3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,27 +27,6 @@ jobs:
       script: make test BUILDTAGS=""
 
     - stage: validate
-#POWERON SUPPORT 
-    - stage: build
-      script: make
-      arch: ppc64le
-    - stage: lint
-      script: make lint BUILDTAGS=""
-      arch: ppc64le
-    - stage: test
-      script: make test BUILDTAGS=""
-      arch: ppc64le
-      go:  1.14.x
-    - stage: test
-      script: make test BUILDTAGS=""
-      arch: ppc64le
-      go: 1.15.x
-    - stage: test
-      script: make test BUILDTAGS=""
-      arch: ppc64le
-      go : tip
-    - stage: validate
-      arch: ppc64le
-
-      script: |
+    
+script: |
         git-validation -run DCO,short-subject -v -range ${TRAVIS_COMMIT_RANGE}
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,9 @@ go:
   - 1.14.x
   - 1.15.x
   - tip
-
+arch:
+  - amd64
+  - ppc64le
 before_install:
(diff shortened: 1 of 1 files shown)


rhatdan:
Please sign your commit.

git commit -a --amend -s
git push --force

asellappen:
> Please sign your commit.
> 
> git commit -a --amend -s
> git push --force

This has been done now.

rhatdan:
LGTM
@kolyshkin @thaJeztah @mrunalp @vrothberg PTAL

.travis.yml
```diff
@@ -25,5 +25,27 @@ jobs:
...
+    - stage: test
+      script: make test BUILDTAGS=""
+      arch: ppc64le
+      go: 1.15.x
+    - stage: test
+      script: make test BUILDTAGS=""
+      arch: ppc64le
+      go : tip
+    - stage: validate
+      arch: ppc64le
```
vrothberg, line 48: I can think we can simplify it more.

The go versions are already specified at the top and Travis will construct the matrix on it's own. So we don't need to specify them here.

In fact, we don't need to add new stages at all but extend the existing ones with:
```
arch:
  - amd64
  - ppc64le
```

asellappen: @vrothberg  , I have tried earlier and it did not work , This was the reason i have specified cpu arch under matrix .

Again i tried as you suggested and ran the jobs and only new arch included in the test stage , I am expecting it should be  included in other stages Validate,build ,lint etc.. , Please find details below 


https://travis-ci.com/github/asellappen/selinux/builds/190412878

Please let me if you have any other thoughts.

asellappen: shall go for the approach which i used in matrix?

.travis.yml
```diff
@@ -25,5 +27,6 @@ jobs:
       script: make test BUILDTAGS=""
 
     - stage: validate
-      script: |
+    
+script: |
```
thaJeztah, line 31: Was this intentional? If I see correctly, this moves the actual script out of the validate step?

thaJeztah (changes requested):

.travis.yml
asellappen: sorry ,it was my mistake , i have correct this move and ran the job, without matrix poweron host is not running for stages build,Validate,Tint. only Test stage is included  power on arch as per below run. 

refer https://travis-ci.com/github/asellappen/selinux for details.

This was the reason i used matrix to run package for both cpu arch.

please check and pass your inputs.

asellappen:
https://github.com/opencontainers/selinux/pull/121 created newly

PR #122, opened 2020-10-22
travis: add ppc to the matrix

kolyshkin:
This is an alternative to #121 with a more concise travis syntax, using the matrix.

Now, git-validation and build-cross are separate things that do no make
sense to have in the matrix, so separate them out to additional jobs.
    
While at it, allow `go: tip` to fail.
    
NOTE that testing with and without selinux tag is handled by the
Makefile, so we do not add this dimension to the matrix.

Closes: #121

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,21 @@
 language: go
 go:
-  - 1.14.x
   - 1.15.x
+  - 1.14.x
   - tip
+arch:
+  - amd64
+  - ppc64le
 
-before_install:
-  - go get -u github.com/vbatts/git-validation
-
-stages:
-  - validate
-  - build
-  - lint
-  - test
+script:
+  - make lint build test
 
 jobs:
+  allow_failures:
+    - go: tip
   include:
-    - stage: build
-      script: make
-
-    - stage: lint
-      script: make lint BUILDTAGS=""
-
-    - stage: test
-      script: make test BUILDTAGS=""
-
     - stage: validate
-      script: |
-        git-validation -run DCO,short-subject -v -range ${TRAVIS_COMMIT_RANGE}
+      before_script: go get -u github.com/vbatts/git-validation
+      script: git-validation -run DCO,short-subject -v -range ${TRAVIS_COMMIT_RANGE}
+    - stage: build-cross
+      script: make build-cross


thaJeztah:
Thanks! Also wanted to check if an alternative approach would work as the other one looked rather verbose 👍

thaJeztah (approved):
LGTM if green 😅

.travis.yml
```diff
@@ -3,27 +3,15 @@ go:
...
-      script: make test BUILDTAGS=""
-
-    - stage: validate
-      script: |
-        git-validation -run DCO,short-subject -v -range ${TRAVIS_COMMIT_RANGE}
+script:
+  - git-validation -run DCO,short-subject -v -range ${TRAVIS_COMMIT_RANGE}
+  - make
+  - make lint BUILDTAGS=""
+  - make test BUILDTAGS=""
```
rhatdan, line 17: Won't this force the tests to disable SELinux?

rhatdan: I know it was there from the beginning.  But it seems like we are testing only code compiled with the selinux disabled flag.
Perhaps we should run it twice. Once with BUILDTAGS="" and once without.

kolyshkin: Right, another dimension to the matrix, let's see if it works.

kolyshkin: fixed!

rhatdan:
LGTM

kolyshkin:
The "validate" job got triplicated for whatever reason, and Travis is super slow today, otherwise I think it's working.

kolyshkin:
Hmm, OK, so maybe we do not need to add BUILDTAGS into matrix, since we already build and test with and without `selinux` tag right from the Makefile.

We should probably just drop the BUILDTAGS from the travis.yml

kolyshkin:
Funny, we also don't run golangci-lint with `selinux` tag enabled as it requires an additional argument to golangci-lint.

ahh

kolyshkin:
> Funny, we also don't run golangci-lint with selinux tag enabled as it requires an additional argument to golangci-lint.

OK, apparently this is not necessary -- warnings generated are the same with and without `--build-tags selinux`

kolyshkin:
Removed buildtags from the travis matrix. Opened https://github.com/opencontainers/selinux/pull/124 to care about golangci-lint,

kolyshkin:
Updated the commit and the PR description.

I still don't entirely grok the Travis CI config syntax, but what I have now is not too far from being good from my POV.

PR #124, opened 2020-10-26
golangci-lint: make sure to run with and without selinux tag

kolyshkin:
... to cover the entire codebase.

Fix a few newly found issues in tests.


vrothberg (approved):
LGTM
@rhatdan PTAL

rhatdan:
LGTM

PR #127, opened 2020-11-06
go-selinux: retry on EINTR

kolyshkin:
Wrap some syscalls (lgetattr, lsetattr, fstatfs, statfs)
to retry on EINTR, just in case.

Related:
 - https://github.com/containers/storage/pull/757
 - https://go-review.googlesource.com/c/go/+/249178
 - https://github.com/golang/go/issues/38033

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -111,7 +111,7 @@ func verifySELinuxfsMount(mnt string) bool {
 		if err == nil {
 			break
 		}
-		if err == unix.EAGAIN {
+		if err == unix.EAGAIN || err == unix.EINTR {
 			continue
 		}
 		return false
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@thaJeztah @vrothberg @saschagrunert @mrunalp PTAL

go-selinux/selinux_linux.go
```diff
@@ -250,8 +250,17 @@ func getSELinuxPolicyRoot() string {
...
-	var buf unix.Statfs_t
-	err := unix.Fstatfs(int(fh.Fd()), &buf)
+	var (
+		buf unix.Statfs_t
+		err error
+	)
+
+	for {
+		err = unix.Fstatfs(int(fh.Fd()), &buf)
+		if err != unix.EINTR {
```
thaJeztah, line 260: I wonder if the approach taken in `verifySELinuxfsMount()` above is somewhat clearer. (I realise it does the same, wondering if that approach causes head-scratching  for the occasional reader).

```go
for {
	err = unix.Fstatfs(int(fh.Fd()), &buf)
	if err == nil {
			break
	}
	if err != unix.EINTR {
		return errors.Wrapf(err, "statfs(%q) failed", fh.Name())
	}
}
```

Or somewhat shorter (not sure if clearer)

```go
for err := unix.Fstatfs(int(fh.Fd()), &buf); err != nil; {
	if err == unix.EINTR {
		continue
	}
	return errors.Wrapf(err, "statfs(%q) failed", fh.Name())
}
```

thaJeztah (commented):
Thanks! Change itself looks good; left some thoughts (happy to get your input, as always)

go-selinux/selinux_linux.go
kolyshkin: Seems that the second version won't work, but the first one, yes, it looks better than the one I wrote this time.

Updated; PTAL

thaJeztah (approved):
LGTM, thanks!

thaJeztah:
LGTM

PR #135, opened 2021-01-04
filePath: fix deadlock

gaurav1086:
Signed-off-by: Gaurav Singh <gaurav1086@gmail.com>

A goroutine returning before calling waitGroup.Done() can cause a deadlock.
Fix this by using defer waitGroup.Done() at the top.

Example: https://play.golang.com/p/n3w0T0wg-6c

### DIFF
--- a/pkg/pwalk/pwalk.go
+++ b/pkg/pwalk/pwalk.go
@@ -55,6 +55,7 @@ func WalkN(root string, walkFn WalkFunc, num int) error {
 	)
 	wg.Add(1)
 	go func() {
+		defer wg.Done()
 		err = filepath.Walk(root, func(p string, info os.FileInfo, err error) error {
 			if err != nil {
 				close(files)
@@ -73,7 +74,6 @@ func WalkN(root string, walkFn WalkFunc, num int) error {
 		if err == nil {
 			close(files)
 		}
-		wg.Done()
 	}()
 
 	wg.Add(num)


kolyshkin:
@gaurav1086 I think you misread the code. There are no returns from the goroutive -- those that you see are from the walker function, which is the last argument of `filepath.Walk`.

IOW the goroutine code is like this:

```go
	go func() {
		defer wg.Done()
		err = filepath.Walk(root, func(p string, info os.FileInfo, err error) error {
			// walker function here
		})
		if err == nil {
			close(files)
		}
		wg.Done()
	}()
```

and a return from that walker function doesn't mean a premature return from the gouroutine.

rhatdan:
@kolyshkin Should we close this?  I think the patch cleans up the code a little to make it easier to comprehend, but I believe there is some penalty to defer calls.

gaurav1086:
> @gaurav1086 I think you misread the code. There are no returns from the goroutive -- those that you see are from the walker function, which is the last argument of `filepath.Walk`.
> 
> IOW the goroutine code is like this:
> 
> ```go
> 	go func() {
> 		defer wg.Done()
> 		err = filepath.Walk(root, func(p string, info os.FileInfo, err error) error {
> 			// walker function here
> 		})
> 		if err == nil {
> 			close(files)
> 		}
> 		wg.Done()
> 	}()
> ```
> 
> and a return from that walker function doesn't mean a premature return from the gouroutine.

@kolyshkin ah yeah, that makes sense. Thanks for clarification.

PR #142, opened 2021-03-30
Free SELinux labels when the labels are not being used

tunacado:
This is a cherry-pick of the following from projectatomic/docker altered
for this version of opencontainers/selinux.

https://github.com/projectatomic/docker/pull/381

Currently when lots of containers are created as disabled, each
container will leak and MCS Label.  Eventually the system will run out
of labels and go into an infinate loop looking for labels.

This also fixes a problem that caused relabeling to occur even when
label=disable was specified, such as with privileged containers.

Signed-off-by: Robb Manes <rmanes@redhat.com>

### DIFF
--- a/go-selinux/label/label_linux.go
+++ b/go-selinux/label/label_linux.go
@@ -47,7 +47,8 @@ func InitLabels(options []string) (plabel string, mlabel string, retErr error) {
 		}
 		for _, opt := range options {
 			if opt == "disable" {
-				return "", mountLabel, nil
+				selinux.ReleaseLabel(processLabel)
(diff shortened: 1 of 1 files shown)


tunacado:
This is the issue described in [Red Hat Bugzilla 1944806](https://bugzilla.redhat.com/show_bug.cgi?id=1944806)

rhatdan:
No we need to have privileged containers content have a label on them, to protect their content from unprivileged containers.

I opened https://github.com/opencontainers/selinux/pull/142/ which will stop wasting MCS Labels for privileged containers.

PR #146, opened 2021-05-17
go.mod: willf/bitset@v1.1.11 -> bits-and-blooms/bitset@v1.2.0

AkihiroSuda:
The repo was renamed: https://github.com/bits-and-blooms/bitset/commit/06a6e5e24e54b3174047e72f1ff1ec5758106812

Changes: https://github.com/bits-and-blooms/bitset/compare/v1.1.11...v1.2.0

PR #147, opened 2021-06-02
Bump to v1.8.2

kolyshkin:
This is to fix the bitset module dependency (#146).

NOTE that instead of using `FutureVersion-dev`, now we use
`CurrentVersion+Dev`, which makes more sense as we can't say
for sure what the future version will be (say 1.8.3 or 1.9.0).


VERSION
```diff
@@ -1 +1 @@
-1.8.2-dev
+1.8.2+dev
```
thaJeztah, line 1: SemVer considers "1.8.2" and "1.8.2+whatever" to be the exact equivalent, so not sure if this is technically correct (if the intent is that "master" is more current than last release) https://semver.org/#spec-item-10

whereas "-dev" denotes a pre-release (so "lower than" the version number) (not great either because -dev > -beta (alphabetically)

kolyshkin: This is what we do in runc. The idea is to mark that this is version 1.8.2 plus some more commits (so SemVer interpretation is more or less correct -- 1.8.2+dev *is* 1.8.2).

I guess as long as this `+dev` suffix won't be used in the actual tag (which we do not plan to do), we're fine.

rhatdan (approved):
LGTM

PR #148, opened 2021-06-11
Stop using pkg/errors, switch to native golang errors, add errorlint and gofumpt

kolyshkin:
* Remove all uses of `pkg/errors`, switching to golang's native error wrapping.
* Replace obsoleted `os.IsPermission` / `os.IsNotExist` with `errors.Is`
* Enable errorlint, fix or suppress its warnings
* Enable gofumpt linter
* Make sure to not return bare unix errors (errno) from public functions.


go-selinux/label/label_linux.go
```diff
@@ -53,11 +54,11 @@ func InitLabels(options []string) (plabel string, mlabel string, retErr error) {
 				return "", selinux.PrivContainerMountLabel(), nil
 			}
 			if i := strings.Index(opt, ":"); i == -1 {
-				return "", "", errors.Errorf("Bad label option %q, valid options 'disable' or \n'user, role, level, type, filetype' followed by ':' and a value", opt)
+				return "", "", fmt.Errorf("Bad label option %q, valid options 'disable' or \n'user, role, level, type, filetype' followed by ':' and a value", opt)
```
thaJeztah, line 57: I hate it that they didn't add a `errors.Errorf()` 😞

go-selinux/selinux_linux.go
```diff
@@ -697,17 +696,21 @@ func socketLabel() (string, error) {
...
 // setKeyLabel takes a process label and tells the kernel to assign the
 // label to the next kernel keyring that gets created
 func setKeyLabel(label string) error {
 	err := writeCon("/proc/self/attr/keycreate", label)
-	if os.IsNotExist(errors.Cause(err)) {
+	if os.IsNotExist(err) {
 		return nil
 	}
-	if label == "" && os.IsPermission(errors.Cause(err)) {
+	if label == "" && os.IsPermission(err) {
```
thaJeztah, line 713: Should we use `errors.Is(err, os.ErrPermission)`?

https://golang.org/pkg/os/#IsPermission

> This function predates errors.Is. It only supports errors returned by the os package. New code should use errors.Is(err, os.ErrPermission).

go-selinux/selinux_linux.go
```diff
@@ -1050,7 +1053,7 @@ func chcon(fpath string, label string, recurse bool) error {
 	return pwalk.Walk(fpath, func(p string, info os.FileInfo, err error) error {
 		e := SetFileLabel(p, label)
 		// Walk a file tree can race with removal, so ignore ENOENT
-		if os.IsNotExist(errors.Cause(e)) {
+		if os.IsNotExist(e) {
```
thaJeztah, line 1056: Same for these?

```suggestion
		if errors.Is(err, os.ErrNotExist) {
```

go-selinux/selinux_linux.go
kolyshkin: Wow, I did not know that the official recommendation is to use `errors.Is` instead of `os.IsXxx`. Will fix.

go-selinux/label/label_linux.go
kolyshkin: Well, could be as easy as `var Errorf = fmt.Errorf` but it's impossible as `fmt` imports `errors` and so `errors` can't import `fmt`.

go-selinux/selinux_linux.go
kolyshkin: fixed

go-selinux/selinux_linux.go
kolyshkin: fixed

go-selinux/selinux_linux.go
kolyshkin: Actually fixed 🤞🏻

kolyshkin:
@thaJeztah PTAL

rhatdan (approved):
LGTM

rhatdan:
@vrothberg @giuseppe PTAL

vrothberg (commented):
LGTM

PR #149, opened 2021-07-30
Release 1.8.3

kolyshkin:
As we have merged #148, we no longer require pkg/errors. It would be nice to have a release cut with this.

Note we're using the same "x.y.z+dev" notation as in 1.8.2 (see https://github.com/opencontainers/selinux/commit/fab6725a74101965f3876f350435b8323fe80167 for rationale).


rhatdan:
LGTM

PR #150, opened 2021-08-06
Add pwalkdir, speed up chcon

kolyshkin:
TL;DR: speed up chcon for up to 1.5x to 2x with Go 1.16+.

1. pkg/pwalkdir: add
    
    This is a copy-paste implementation of pwalk but utilizing
    filepath.WalkDir instead (added in Go 1.16). As WalkDir do not call
    stat(2) on every entry, it is much (2x on my machine) faster than Walk
    (see below).
    
    The idea is to have pwalk and pwalkdir coexist for some time, then
    gradually retire pwalk once Go < 1.16 is unsupported.

2. Chcon: speedup for Go 1.16+
    
    Using the newly added pwalkdir, the benchmark added by and described
    in commit ba956cafe09661021bf shows a considerable speed improvement
    (0.8s with pwalk, 0.5s with pwalkdir, 0.9s with chcon from GNU coreutils).

### DIFF
--- a/pkg/pwalkdir/pwalkdir.go
+++ b/pkg/pwalkdir/pwalkdir.go
@@ -1,0 +1,103 @@
+// +build go1.16
+
+package pwalkdir
+
+import (
+	"fmt"
+	"io/fs"
+	"path/filepath"
+	"runtime"
+	"sync"
+)
+
+// Walk is a wrapper for filepath.WalkDir which can call multiple walkFn
+// in parallel, allowing to handle each item concurrently. A maximum of
+// twice the runtime.NumCPU() walkFn will be called at any one time.
+// If you want to change the maximum, use WalkN instead.
+//
+// The order of calls is non-deterministic.
+//
+// Note that this implementation only supports primitive error handling:
+//
+// - no errors are ever passed to walkFn;
+//
+// - once a walkFn returns any error, all further processing stops
+// and the error is returned to the caller of Walk;
+//
+// - filepath.SkipDir is not supported;
+//
+// - if more than one walkFn instance will return an error, only one
+// of such errors will be propagated and returned by Walk, others
+// will be silently discarded.
+func Walk(root string, walkFn fs.WalkDirFunc) error {
+	return WalkN(root, walkFn, runtime.NumCPU()*2)
+}
+
+// WalkN is a wrapper for filepath.WalkDir which can call multiple walkFn
+// in parallel, allowing to handle each item concurrently. A maximum of
+// num walkFn will be called at any one time.
+//
+// Please see Walk documentation for caveats of using this function.
+func WalkN(root string, walkFn fs.WalkDirFunc, num int) error {
+	// make sure limit is sensible
+	if num < 1 {
+		return fmt.Errorf("walk(%q): num must be > 0", root)
+	}
+
+	files := make(chan *walkArgs, 2*num)
+	errCh := make(chan error, 1) // Get the first error, ignore others.
+
+	// Start walking a tree asap.
+	var (
+		err error
+		wg  sync.WaitGroup
+	)
+	wg.Add(1)
+	go func() {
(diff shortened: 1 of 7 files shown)


kolyshkin:
A few non-scientific benchmarks (testdata is Linux kernel source tree, find is for reference only)

| Op | Time |
| --- | --- |
| find -print >/dev/null | 0.19s |
| chcon -R | 0.92s |
| Before this PR (pwalk) | 0.83s |
| This PR (pwalkdir) + go 1.16 | 0.51s |
| This PR (pwalkdir) + go 1.17rc2 | 0.45s |

thaJeztah:
Nice! Will try to do a review soon, but ping me otherwise 😅

rhatdan (approved):
LGTM

rhatdan:
Awesome work @kolyshkin

pkg/pwalkdir/pwalkdir.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"fmt"
+	"io/fs"
+	"path/filepath"
+	"runtime"
+	"sync"
+)
+
+type WalkFunc = fs.WalkDirFunc
```
thaJeztah, line 13: Probably want to rename the alias as well. `WalkDirFunc` is not a drop-in replacement for `pwalk.WalkFunc` (different signature), so perhaps better to rename?

```suggestion
type WalkDirFunc = fs.WalkDirFunc
```

thaJeztah: Or should we remove the alias altogether? The signature has `fs.Direntry` in it, so consumers would still have to add `import "io/fs"` (so no extra convenience there)

thaJeztah (changes requested):
Thanks! Spotted a typo, and one question/suggestion (rename or drop the `WalkFunc` type alias); let me know what you think

pkg/pwalkdir/pwalkdir.go
kolyshkin: Yep, I think it is best to remove the alias.

kolyshkin: removed

pkg/pwalkdir/pwalkdir.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Walk is a wrapper for filepath.WalkDir which can call multiple walFn
```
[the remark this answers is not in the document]
kolyshkin: fixed

thaJeztah:
Looks like you forgot to remove some references that used the alias;

```
  level=error msg="Running error: buildir: analysis skipped: errors in package: [/home/runner/work/selinux/selinux/pkg/pwalkdir/pwalkdir_test.go:124:42: undeclared name: WalkFunc /home/runner/work/selinux/selinux/pkg/pwalkdir/pwalkdir_test.go:127:34: undeclared name: WalkFunc /home/runner/work/selinux/selinux/pkg/pwalkdir/pwalkdir_test.go:141:8: undeclared name: WalkFunc /home/runner/work/selinux/selinux/pkg/pwalkdir/pwalkdir_test.go:153:24: cannot use filepath.WalkDir (value of type func(root string, fn fs.WalkDirFunc) error) as walkerFunc value in struct literal /home/runner/work/selinux/selinux/pkg/pwalkdir/pwalkdir_test.go:154:21: cannot use Walk (value of type func(root string, walkFn fs.WalkDirFunc) error) as walkerFunc value in struct literal]"
```

kolyshkin:
> Looks like you forgot to remove some references that used the alias

🤦🏻 never trust 3am edits! Fixed.

thaJeztah:
> 🤦🏻 never trust 3am edits! Fixed

More than familiar with those 😂

thaJeztah (approved):
LGTM!

thaJeztah:
@rhatdan ptal

PR #151, opened 2021-08-07
pkg/pwalk: fix minor docs issues

thaJeztah:
spotted these while reviewing https://github.com/opencontainers/selinux/pull/150, which had these fixed in the new implementation


thaJeztah:
@kolyshkin ptal

kolyshkin (approved):
The plan is to retire pwalk once Go < 1.16 is not supported, this is why I haven't backported those minor fixes.

OTOH they won't hurt.

LGTM

rhatdan:
LGTM

PR #153, opened 2021-09-07
Lazy init

kolyshkin:
With the following proggy:
```go
package main

import 	_ "github.com/opencontainers/selinux/go-selinux"

func main() {}
```

we can easily measure the overhead of merely importing (not actually using) the package:

```console
$ GODEBUG=inittrace=1 go run a.go 2>&1 | grep go-selinux
# github.com/opencontainers/selinux/go-selinux
init github.com/opencontainers/selinux/go-selinux @0.87 ms, 0.15 ms clock, 18328 bytes, 152 allocs
```

This PR tries to minimize that by:
* removing regex usage;
* initializing some stuff in a lazy (on demand) manner, rather than during initialization time

With these patches, we have
```console
init github.com/opencontainers/selinux/go-selinux @0.53 ms, 0 ms clock, 160 bytes, 8 allocs
```

which eliminates most of the overhead.

Besides, modified `readConfig` and `loadLabels` show some performance and memory improvements as well:
```console
[kir@kir-rhat go-selinux]$ benchstat before after
name          old time/op    new time/op    delta
ReadConfig-4    4.86µs ± 1%    3.46µs ± 2%  -28.82%  (p=0.002 n=6+6)
LoadLabels-4    12.7µs ± 3%     6.8µs ± 1%  -46.24%  (p=0.002 n=6+6)

name          old alloc/op   new alloc/op   delta
ReadConfig-4    5.08kB ± 0%    4.35kB ± 0%  -14.33%  (p=0.002 n=6+6)
LoadLabels-4    6.49kB ± 0%    6.10kB ± 0%   -6.07%  (p=0.000 n=6+5)

name          old allocs/op  new allocs/op  delta
ReadConfig-4      20.0 ± 0%       8.0 ± 0%  -60.00%  (p=0.002 n=6+6)
LoadLabels-4      45.0 ± 0%      45.0 ± 0%     ~     (all equal)
```

See individual commit messages for details.

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -81,6 +81,10 @@ var (
 	// for policyRoot()
 	policyRootOnce sync.Once
 	policyRootVal  string
+
+	// for label()
+	loadLabelsOnce sync.Once
+	labels         map[string]string
 )
 
 func policyRoot() string {
@@ -903,13 +907,11 @@ func openContextFile() (*os.File, error) {
 	return os.Open(filepath.Join(policyRoot(), "/contexts/lxc_contexts"))
 }
 
-var labels, privContainerMountLabel = loadLabels()
-
-func loadLabels() (map[string]string, string) {
-	labels := make(map[string]string)
+func loadLabels() {
+	labels = make(map[string]string)
 	in, err := openContextFile()
 	if err != nil {
-		return labels, ""
+		return
 	}
 	defer in.Close()
 
@@ -933,30 +935,37 @@ func loadLabels() (map[string]string, string) {
 
 	con, _ := NewContext(labels["file"])
 	con["level"] = fmt.Sprintf("s0:c%d,c%d", maxCategory-2, maxCategory-1)
-	reserveLabel(con.get())
-	return labels, con.get()
+	privContainerMountLabel = con.get()
+	reserveLabel(privContainerMountLabel)
+}
+
+func label(key string) string {
+	loadLabelsOnce.Do(func() {
+		loadLabels()
+	})
+	return labels[key]
 }
 
 // kvmContainerLabels returns the default processLabel and mountLabel to be used
 // for kvm containers by the calling process.
 func kvmContainerLabels() (string, string) {
-	processLabel := labels["kvm_process"]
+	processLabel := label("kvm_process")
 	if processLabel == "" {
-		processLabel = labels["process"]
+		processLabel = label("process")
 	}
 
-	return addMcs(processLabel, labels["file"])
+	return addMcs(processLabel, label("file"))
 }
 
 // initContainerLabels returns the default processLabel and file labels to be
 // used for containers running an init system like systemd by the calling process.
 func initContainerLabels() (string, string) {
-	processLabel := labels["init_process"]
+	processLabel := label("init_process")
 	if processLabel == "" {
-		processLabel = labels["process"]
+		processLabel = label("process")
 	}
 
-	return addMcs(processLabel, labels["file"])
+	return addMcs(processLabel, label("file"))
 }
 
 // containerLabels returns an allocated processLabel and fileLabel to be used for
@@ -966,9 +975,9 @@ func containerLabels() (processLabel string, fileLabel string) {
 		return "", ""
 	}
 
-	processLabel = labels["process"]
-	fileLabel = labels["file"]
-	readOnlyFileLabel = labels["ro_file"]
+	processLabel = label("process")
+	fileLabel = label("file")
+	readOnlyFileLabel = label("ro_file")
 
 	if processLabel == "" || fileLabel == "" {
 		return "", fileLabel
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -12,7 +12,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -68,7 +67,6 @@ const (
 )
 
 var (
-	assignRegex       = regexp.MustCompile(`^([^=]+)=(.*)$`)
 	readOnlyFileLabel string
 	state             = selinuxState{
 		mcsList: make(map[string]bool),
@@ -236,7 +234,7 @@ func readConfig(target string) string {
 	scanner := bufio.NewScanner(in)
 
 	for scanner.Scan() {
-		line := strings.TrimSpace(scanner.Text())
+		line := bytes.TrimSpace(scanner.Bytes())
 		if len(line) == 0 {
 			// Skip blank lines
 			continue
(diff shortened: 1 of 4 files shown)


vrothberg (commented):
LGTM, nice catch

go-selinux/selinux_linux.go
```diff
@@ -935,12 +944,13 @@ func loadLabels() (map[string]string, string) {
 	con, _ := NewContext(labels["file"])
 	con["level"] = fmt.Sprintf("s0:c%d,c%d", maxCategory-2, maxCategory-1)
 	reserveLabel(con.get())
-	return labels, con.get()
+	privContainerMountLabel = con.get()
```
thaJeztah, line 948: If we're optimising; I see we're calling `con.get()` twice here.  Looks like we can swap the order;

```go
privContainerMountLabel = con.get()
reserveLabel(privContainerMountLabel)
```

go-selinux/selinux_linux.go
```diff
@@ -935,12 +944,13 @@ func loadLabels() (map[string]string, string) {
...
 }
 
 // kvmContainerLabels returns the default processLabel and mountLabel to be used
 // for kvm containers by the calling process.
 func kvmContainerLabels() (string, string) {
+	loadLabels()
 	processLabel := labels["kvm_process"]
 	if processLabel == "" {
```
thaJeztah, line 956: I see most (all?) places we call `loadLabels()` are followed by `something := labels[...]`. Having to call `loadLabels()` is something that _could_ be forgotten.

What if we make this more "natural", and instead of calling `loadLabels()` directly, implement it as an accessor for `labels`, something like:

```go
func getLabel(name string) string {
	loadLabelsOnce.Do(func() {
		doLoadLabels()
	})
	return labels[name]
}
```

Then this could be written as

```suggestion
	processLabel := getLabel("kvm_process")
  	if processLabel == "" {
```

go-selinux/selinux_linux.go
```diff
@@ -237,20 +235,21 @@ func readConfig(target string) string {
 	scanner := bufio.NewScanner(in)
 
 	for scanner.Scan() {
-		line := strings.TrimSpace(scanner.Text())
-		if len(line) == 0 {
+		str := bytes.TrimSpace(scanner.Bytes())
```
thaJeztah, line 238: Was there a reason to rename `line` to `str`? (it's slightly confusing because `str` is not a _string_ but a `[]byte` now)

go-selinux/selinux_linux.go
```diff
@@ -237,20 +235,21 @@ func readConfig(target string) string {
...
+		if str[0] == ';' || str[0] == '#' {
 			// Skip comments
 			continue
 		}
-		if groups := assignRegex.FindStringSubmatch(line); groups != nil {
-			key, val := strings.TrimSpace(groups[1]), strings.TrimSpace(groups[2])
-			if key == target {
-				return strings.Trim(val, "\"")
-			}
+		fields := bytes.SplitN(str, []byte{'='}, 2)
```
thaJeztah, line 247: Just noticing this: this also accepts `=` inside the value part;  this was already the case in the regex, but mostly wondering what values are accepted (white-space included if there's no quotes? `=` included?)

Not for this PR, but perhaps something that needs to be checked (in a follow-up)

go-selinux/selinux_linux.go
```diff
@@ -926,19 +925,21 @@ func doLoadLabels() {
 	scanner := bufio.NewScanner(in)
 
 	for scanner.Scan() {
-		line := strings.TrimSpace(scanner.Text())
-		if len(line) == 0 {
+		str := bytes.TrimSpace(scanner.Bytes())
```
thaJeztah, line 928: same question here for `line -> str`

go-selinux/selinux_linux.go
```diff
@@ -79,8 +77,20 @@ var (
...
 	haveThreadSelf bool
+
+	// for getPolicyRoot
+	getPolicyRootOnce sync.Once
+	policyRoot        string
 )
 
+func getPolicyRoot() string {
+	getPolicyRootOnce.Do(func() {
+		policyRoot = filepath.Join(selinuxDir, readConfig(selinuxTypeTag))
```
thaJeztah, line 91: I noticed that `readConfig` is also called in `defaultEnforceMode()`. Wondering if instead of applying a `sync.Once` here, we should make `readConfig` store the results as a map (similar to `doReadLabels()`) and have a sync.Once for that 🤔 

Not sure if that makes sense (mostly thinking out loud as usual)

thaJeztah (changes requested):
left some questions / suggestions / thoughts 😅

rhatdan:
LGTM, but need to answer @thaJeztah feedback.

go-selinux/selinux_linux.go
kolyshkin: Yes, makes sense (except the `PrivContainerMountLabel` looks a tad odd, but I've added a comment).

go-selinux/selinux_linux.go
kolyshkin: We're using `SplitN` so only the first (i.e. leftmost) `=` is used for splitting. Now, variable name can not contain `=` so we're good here.

Whitespace is not accepted (around `=`) because the same file is also parsed by shell, which does not allow spaces around `=`.

go-selinux/selinux_linux.go
kolyshkin: I thought about it as well (first time a year or so ago when I saw the code for the first time). This is a small text file, only containing two variables, and `DefaultEnforceMode` is (I guess) rarely used, so it's not worth optimizing for.

go-selinux/selinux_linux.go
```diff
@@ -81,6 +81,11 @@ var (
 	// for getPolicyRoot
 	getPolicyRootOnce sync.Once
 	policyRoot        string
+
+	// for loadLabels
+	labels                  map[string]string
+	privContainerMountLabel string
+	loadLabelsOnce          sync.Once
```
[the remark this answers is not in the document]
kolyshkin: 1. done
2. I'd like it to stay as is, because otherwise it would be less clear that `privContainerMountLabel` is only initialized together with `labels` (I _think_ that we can't move that to `PrivContainerMountLabel()` because of the reservation).

go-selinux/selinux_linux.go
kolyshkin: Good catch! Done

go-selinux/selinux_linux.go
kolyshkin: Yes, that was not called for, and introduces some additional changes that are not needed.

Reverted back.

go-selinux/selinux_linux.go
kolyshkin: fixed

go-selinux/selinux_linux.go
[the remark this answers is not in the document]
kolyshkin: Re: 2. OTOH the code is not very complex and it is easy to figure out how `privContainerMountLabel` is initialized.

Done.

go-selinux/selinux_linux.go
thaJeztah: Yeah, we can probably continue to just assume "whatever comes after `=` (minus quotes) _should_ be valid". It was mostly an observation.

thaJeztah (approved):
LGTM

thanks! it's both cleaner _and_ more performant, which is always nice!

thaJeztah:
@rhatdan ptal; ready to go if it still LGTY

rhatdan:
Rebase and we can merge, and I will cut a new release.

kolyshkin:
rebased

thaJeztah (approved):
still LGTM, thanks!

rhatdan (approved):
LGTM

PR #154, opened 2021-09-08
(Context).get: speedup

kolyshkin:
`fmt.Sprintf` is slow. Let's use string concatenation instead, and avoid
looking up level label twice.

Benchmark shows 3x improvement in performance, and less allocations.
```
name          old time/op    new time/op    delta
ContextGet-4     277ns ± 1%      93ns ± 1%  -66.55%  (p=0.016 n=4+5)

name          old alloc/op   new alloc/op   delta
ContextGet-4      128B ± 0%       64B ± 0%     ~     (p=0.079 n=4+5)

name          old allocs/op  new allocs/op  delta
ContextGet-4      5.00 ± 0%      1.00 ± 0%     ~     (p=0.079 n=4+5)
```
Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -723,10 +723,10 @@ func keyLabel() (string, error) {
 
 // get returns the Context as a string
 func (c Context) get() string {
-	if c["level"] != "" {
-		return fmt.Sprintf("%s:%s:%s:%s", c["user"], c["role"], c["type"], c["level"])
+	if level := c["level"]; level != "" {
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
I am not sure this is called that often.

rhatdan (approved):

rhatdan (approved):

kolyshkin:
> I am not sure this is called that often.

It's not, but the fix is trivial, too.

thaJeztah (approved):
LGTM

PR #155, opened 2021-09-08
readCon speedup

kolyshkin:
When looking at runc strace output, I noticed that selinux attr files
are read byte by byte:

	read(8, "u", 1) = 1
	read(8, "n", 1) = 1
	....

Apparently, this happens because fmt.Fscanf("%s") works that way.

Let's read the whole file via ioutil.ReadAll instead.

The added benchmark shows a drastic improvement:
```
name            old time/op    new time/op    delta
CurrentLabel-4    26.4µs ± 1%     4.3µs ± 1%   -83.73%  (p=0.004 n=6+5)

name            old alloc/op   new alloc/op   delta
CurrentLabel-4      352B ± 0%      776B ± 0%  +120.45%  (p=0.002 n=6+6)

name            old allocs/op  new allocs/op  delta
CurrentLabel-4      9.00 ± 0%      7.00 ± 0%   -22.22%  (p=0.002 n=6+6)

Note that increased allocation size is caused by the fact that ReadAll
preallocates a 512 bytes buffer, which is very reasonable.
```

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -274,12 +274,15 @@ func readCon(fpath string) (string, error) {
 	if err := isProcHandle(in); err != nil {
 		return "", err
 	}
+	return readConFd(in)
+}
 
-	var retval string
-	if _, err := fmt.Fscanf(in, "%s", &retval); err != nil {
+func readConFd(in *os.File) (string, error) {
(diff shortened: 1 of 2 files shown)


rhatdan (approved):
LGTM

vrothberg (commented):
LGTM

PR #156, opened 2021-09-08
writeCon: no need to wrap os.Write error

kolyshkin:
The errors returned from os are usually descriptive enough
(i.e. they contain the file name and the failed operation).
    
Fixes: bb3d708b

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -389,7 +389,7 @@ func writeCon(fpath, val string) error {
 		_, err = out.Write(nil)
 	}
 	if err != nil {
-		return &os.PathError{Op: "write", Path: fpath, Err: err}
+		return err
 	}
 	return nil
 }


rhatdan (approved):
LGTM

go-selinux/selinux_linux.go
```diff
@@ -389,7 +389,7 @@ func writeCon(fpath, val string) error {
 		_, err = out.Write(nil)
 	}
 	if err != nil {
-		return &os.PathError{Op: "write", Path: fpath, Err: err}
+		return err
 	}
 	return nil
```
thaJeztah, line 394: Surprised the linter didn't complain that this could just `return err` (which could be nil, or not).

I'm ok with the current version as the linter is ok with it (and it may be somewhat clearer)

thaJeztah (approved):
LGTM

PR #157, opened 2021-09-09
Add the list of container read only file types

rhatdan:
These are the types of files that container domains are allowed to read.
We want this for contianer storage, to allow us to do some deduplicating
of SELinux types for use within container/storage.

This list is very static and has been basically the same for years.
Generating this programatically, while possible is very costly.


rhatdan:
@giuseppe @mtrmac PTAL

rhatdan:
I think we will just add these directly to containers.storage and allow it to be configured in storage.conf.

PR #160, opened 2021-09-24
go-selinux: add a Maybe option for Relabel and Chcon

haircommander:
When Maybe{Relabel,Chcon} are called, the path they're called for
is only relabeled when the top-level is not already labelled.

Signed-off-by: Peter Hunt <pehunt@redhat.com>

### DIFF
--- a/go-selinux/label/label_linux.go
+++ b/go-selinux/label/label_linux.go
@@ -103,10 +103,23 @@ func SetFileCreateLabel(fileLabel string) error {
 	return selinux.SetFSCreateLabel(fileLabel)
 }
 
-// Relabel changes the label of path to the filelabel string.
+// Relabel changes the label of path and all the entries beneath the path,
 // It changes the MCS label to s0 if shared is true.
 // This will allow all containers to share the content.
 func Relabel(path string, fileLabel string, shared bool) error {
+	return relabel(path, fileLabel, shared, false)
+}
+
+// MaybeRelabel changes the label of path and all the entries beneath the path,
+// It changes the MCS label to s0 if shared is true.
+// This will allow all containers to share the content.
+// Unlike Relabel, the relabeling is only performed if path's label do not match the requested one.
+func MaybeRelabel(path string, fileLabel string, shared bool) error {
+	return relabel(path, fileLabel, shared, true)
+}
+
+// An internal function that relabels a path.
+func relabel(path string, fileLabel string, shared, stopIfAlreadyLabeled bool) error {
 	if !selinux.GetEnabled() || fileLabel == "" {
 		return nil
 	}
@@ -164,10 +177,10 @@ func Relabel(path string, fileLabel string, shared bool) error {
 		c["level"] = "s0"
 		fileLabel = c.Get()
 	}
-	if err := selinux.Chcon(path, fileLabel, true); err != nil {
-		return err
+	if stopIfAlreadyLabeled {
+		return selinux.MaybeChcon(path, fileLabel, true)
 	}
-	return nil
+	return selinux.Chcon(path, fileLabel, true)
 }
 
 // DisableSecOpt returns a security opt that can disable labeling
--- a/go-selinux/selinux.go
+++ b/go-selinux/selinux.go
@@ -256,7 +256,15 @@ func CopyLevel(src, dest string) (string, error) {
 // If fpath is a directory and recurse is true, then Chcon walks the
 // directory tree setting the label.
 func Chcon(fpath string, label string, recurse bool) error {
-	return chcon(fpath, label, recurse)
+	return chcon(fpath, label, recurse, false)
+}
+
+// MaybeChcon changes the fpath file object to the SELinux label label.
+// If fpath is a directory and recurse is true, then Chcon walks the
+// directory tree setting the label, as long as fpath was not already labeled.
+// with the requested label.
+func MaybeChcon(fpath string, label string, recurse bool) error {
+	return chcon(fpath, label, recurse, true)
 }
 
 // DupSecOpt takes an SELinux process label and returns security options that
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -1050,8 +1050,10 @@ func badPrefix(fpath string) error {
 
 // chcon changes the fpath file object to the SELinux label label.
 // If fpath is a directory and recurse is true, then chcon walks the
-// directory tree setting the label.
-func chcon(fpath string, label string, recurse bool) error {
+// directory tree setting the label, except when stopIfAlreadyLabeled is true and
+// fpath is already labeled correctly, in which case the relabeling
+// is not performed.
+func chcon(fpath string, label string, recurse, stopIfAlreadyLabeled bool) error {
 	if fpath == "" {
 		return ErrEmptyPath
 	}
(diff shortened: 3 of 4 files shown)


haircommander:
This has not really been tested yet, so I think it should be considered a WIP for now

rhatdan:
This Looks fine with me.  It could end up with users hitting weird errors, where the tree was not fully labeled.  But with the issues Kubernetes is having with relabeling, this seems like a viable option.

haircommander:
yeah, our plan is to make sure it's optin (with an annotation on the pod that wants to use it) and is documented as USE AT OWN RISK very clearly

kolyshkin:
I like the idea and the implementation, except for the `Lazy` prefix. For me, lazy means doing something on demand (in this particular case, the file would be relabeled not now, but some time later, right before it's first accessed ... not that such functionality exist).

Can I propose `CondRelabel` (for "conditional") or `MaybeRelabel`?

go-selinux/label/label_linux.go
```diff
@@ -107,6 +107,19 @@ func SetFileCreateLabel(fileLabel string) error {
 // It changes the MCS label to s0 if shared is true.
 // This will allow all containers to share the content.
 func Relabel(path string, fileLabel string, shared bool) error {
+	return relabel(path, fileLabel, shared, false)
+}
+
+// LazyRelabel changes the label of path to the filelabel string.
+// It changes the MCS label to s0 if shared is true.
+// This will allow all containers to share the content.
+// It only does so if "path" does not already have the requested label.
```
kolyshkin, line 116:
```suggestion
Unlike Relabel, the relabeling is only performed if path's label do not match the requested one.
```

go-selinux/label/label_linux.go
```diff
@@ -107,6 +107,19 @@ func SetFileCreateLabel(fileLabel string) error {
 // It changes the MCS label to s0 if shared is true.
 // This will allow all containers to share the content.
 func Relabel(path string, fileLabel string, shared bool) error {
+	return relabel(path, fileLabel, shared, false)
+}
+
+// LazyRelabel changes the label of path to the filelabel string.
```
kolyshkin, line 113:
```suggestion
// LazyRelabel changes the label of path and, recursively, all the entries beneath the path,
// to that specified by the fileLabel argument.
```

kolyshkin: or somesuch...

go-selinux/selinux.go
```diff
@@ -252,11 +252,19 @@ func CopyLevel(src, dest string) (string, error) {
 	return copyLevel(src, dest)
 }
 
+// LazyChcon changes the fpath file object to the SELinux label label.
+// If fpath is a directory and recurse is true, then Chcon walks the
+// directory tree setting the label, as long as fpath was not already labelled
```
kolyshkin, line 257: British English detected :)

nit: `labeled`

go-selinux/selinux_linux.go
```diff
@@ -1050,8 +1050,10 @@ func badPrefix(fpath string) error {
 
 // chcon changes the fpath file object to the SELinux label label.
 // If fpath is a directory and recurse is true, then chcon walks the
-// directory tree setting the label.
-func chcon(fpath string, label string, recurse bool) error {
+// directory tree setting the label in most cases.
+// When lazy is true, if fpath is already labelled correctly, the
```
kolyshkin, line 1054:
```suggestion
// directory tree setting the label, except when lazy is true and fpath is already
// labeled correctly, in which case the relabeling is not peformed.
```

or somesuch

go-selinux/selinux_linux.go
```diff
@@ -1066,6 +1068,18 @@ func chcon(fpath string, label string, recurse bool) error {
 		return setFileLabel(fpath, label)
 	}
 
+	if lazy {
+		currentLabel, err := fileLabel(fpath)
+		if err != nil {
+			return err
+		}
+		// Top level is already labelled correctly,
```
kolyshkin, line 1076: nit: labelled -> labeled

rhatdan:
I agree, when I first saw this PR, with the name lazy, I thought it was just going to return success and then run a thread in the background to label the volume.

I instantly thought about

```
man 2 umount
...
      MNT_DETACH (since Linux 2.4.11)
              Perform  a  lazy unmount: make the mount point unavailable for new accesses, immediately disconnect the filesystem and all filesystems mounted below it from each other and from the mount table, and actually perform the unmount when the mount point ceases to be busy.
```
BTW Would this be possible a flag to say check the top layer, and return success if it is labeled correcly, but also spin off a thread the walk the file system fixing labels, and then just logrus the error if one happens.

go-selinux/selinux_linux.go
```diff
@@ -1066,6 +1068,18 @@ func chcon(fpath string, label string, recurse bool) error {
 		return setFileLabel(fpath, label)
 	}
 
+	if lazy {
+		currentLabel, err := fileLabel(fpath)
```
tjungblu, line 1072: @haircommander do we have a guarantee that the top level will always be labeled last? Especially with `pwalkdir`.

haircommander: not yet, though @kolyshkin is going to work on such an enhancement

haircommander:
comments addressed, I switched to `Maybe`

go-selinux/selinux_stub.go
```diff
@@ -135,7 +135,7 @@ func copyLevel(src, dest string) (string, error) {
 	return "", nil
 }
 
-func chcon(fpath string, label string, recurse bool) error {
+func chcon(fpath string, label string, recurse, lazy bool) error {
```
kolyshkin, line 138: `s/lazy/conditional/`?

go-selinux/label/label_linux.go
```diff
@@ -103,10 +103,23 @@ func SetFileCreateLabel(fileLabel string) error {
...
+// MaybeRelabel changes the label of path and all the entries beneath the path,
+// It changes the MCS label to s0 if shared is true.
+// This will allow all containers to share the content.
+// Unlike Relabel, the relabeling is only performed if path's label do not match the requested one.
+func MaybeRelabel(path string, fileLabel string, shared bool) error {
+	return relabel(path, fileLabel, shared, true)
+}
+
+// An internal function that relabels a path.
+func relabel(path string, fileLabel string, shared, lazy bool) error {
```
kolyshkin, line 122: I would not use `lazy` here either.

kolyshkin:
PR subject and description need to be updated accordingly (Lazy -> Maybe)

haircommander:
fixed!

kolyshkin (changes requested):
After a fresh look, maybe we'd better not add this to the library, instead implementing this check in the caller (provided that https://github.com/opencontainers/selinux/pull/161 is in).

go-selinux/selinux_linux.go
kolyshkin: Please see https://github.com/opencontainers/selinux/pull/161

kolyshkin:
> After a fresh look, maybe we'd better not add this to the library, instead implementing this check in the caller (provided that #161 is in).

IOW the caller can check the top-level path label, and call relabel conditionally by itself.

haircommander:
closing this, because it can be handled in caller

PR #161, opened 2021-09-27
Chcon,Relabel: process the argument last

kolyshkin:
Modify pwalk and pwalkdir to make sure the path itself is added last.

This adds a guarantee that Chcon/Relabel will relabel the top-level
directory (i.e. the argument) only after all its descendants are
relabeled, thus creating a possibility to do conditional relabeling.

Benchmarks showed no noticeable difference in performance.


rhatdan (approved):
LGTM

kolyshkin:
@thaJeztah PTAL

thaJeztah (approved):
LGTM

PR #165, opened 2021-10-06
Relabel,Chcon: do process the root entry last

kolyshkin:
Commit 8501cfdf (PR #161) does not guarantee that the root entry
is processed last -- due to parallel nature of the pwalk it one of the
last to be processed, but might be not the last one.
    
Rewrite the code to process the root entry after all the others
are done.
    
Remove the test for root entry being last as the `last = p` assignment
is racy, and this is now very obvious from the code that the root entry
is indeed the last one.
    
Fixes: 8501cfdfaeec1072e7e 🤦🏻 

While at it:
 * add go 1.17 to test matrix
 * also test with `-race` flag

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -51,6 +51,7 @@ jobs:
       fail-fast: false
       matrix:
         go-version: [1.15.x, 1.16.x, 1.17.x]
+        race: ["-race", ""]
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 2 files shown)


kolyshkin:
Once merged, we need to
* add the new tests (1.17.x and -race variants) to required
* release 1.9.1

rhatdan (approved):
LGTM

haircommander:
LGTM

PR #169, opened 2021-11-16
Add LFileLabel and LSetFileLabel

rhatdan:
SELinux C library has two functions for dealing with file labels, one
which follows symlinks and one that does not.

Golang bindings should work the same way.  The lack of this function is
resulting in https://github.com/containers/buildah/pull/3630 which has
to hack around the problem.

### DIFF
--- a/go-selinux/selinux.go
+++ b/go-selinux/selinux.go
@@ -61,14 +61,28 @@ func ClassIndex(class string) (int, error) {
 	return classIndex(class)
 }
 
-// SetFileLabel sets the SELinux label for this path or returns an error.
+// SetFileLabel sets the SELinux label for this path, following symlinks,
+// or returns an error.
 func SetFileLabel(fpath string, label string) error {
 	return setFileLabel(fpath, label)
 }
 
-// FileLabel returns the SELinux label for this path or returns an error.
+// LsetFileLabel sets the SELinux label for this path, not following symlinks,
+// or returns an error.
+func LsetFileLabel(fpath string, label string) error {
+	return lSetFileLabel(fpath, label)
+}
+
+// FileLabel returns the SELinux label for this path, following symlinks,
+// or returns an error.
 func FileLabel(fpath string) (string, error) {
 	return fileLabel(fpath)
+}
+
+// LfileLabel returns the SELinux label for this path, not following symlinks,
+// or returns an error.
+func LfileLabel(fpath string) (string, error) {
+	return lFileLabel(fpath)
 }
 
 // SetFSCreateLabel tells the kernel what label to use for all file system objects
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -316,8 +316,9 @@ func classIndex(class string) (int, error) {
 	return index, nil
 }
 
-// setFileLabel sets the SELinux label for this path or returns an error.
-func setFileLabel(fpath string, label string) error {
+// lSetFileLabel sets the SELinux label for this path, not following symlinks,
+// or returns an error.
+func lSetFileLabel(fpath string, label string) error {
 	if fpath == "" {
 		return ErrEmptyPath
 	}
@@ -334,8 +335,46 @@ func setFileLabel(fpath string, label string) error {
 	return nil
 }
 
-// fileLabel returns the SELinux label for this path or returns an error.
+// setFileLabel sets the SELinux label for this path, following symlinks,
+// or returns an error.
+func setFileLabel(fpath string, label string) error {
+	if fpath == "" {
+		return ErrEmptyPath
+	}
+	for {
+		err := unix.Setxattr(fpath, xattrNameSelinux, []byte(label), 0)
+		if err == nil {
+			break
+		}
+		if err != unix.EINTR { //nolint:errorlint // unix errors are bare
+			return &os.PathError{Op: "setxattr", Path: fpath, Err: err}
+		}
+	}
+
+	return nil
+}
+
+// fileLabel returns the SELinux label for this path, following symlinks,
+// or returns an error.
 func fileLabel(fpath string) (string, error) {
+	if fpath == "" {
+		return "", ErrEmptyPath
+	}
+
+	label, err := getxattr(fpath, xattrNameSelinux)
+	if err != nil {
+		return "", &os.PathError{Op: "getxattr", Path: fpath, Err: err}
+	}
+	// Trim the NUL byte at the end of the byte buffer, if present.
+	if len(label) > 0 && label[len(label)-1] == '\x00' {
+		label = label[:len(label)-1]
+	}
+	return string(label), nil
+}
+
+// lFileLabel returns the SELinux label for this path, not following symlinks,
+// or returns an error.
+func lFileLabel(fpath string) (string, error) {
 	if fpath == "" {
 		return "", ErrEmptyPath
 	}
--- a/go-selinux/xattrs_linux.go
+++ b/go-selinux/xattrs_linux.go
@@ -36,3 +36,36 @@ func doLgetxattr(path, attr string, dest []byte) (int, error) {
 		}
 	}
 }
+
+// getxattr returns a []byte slice containing the value of
+// an extended attribute attr set for path.
+func getxattr(path, attr string) ([]byte, error) {
+	// Start with a 128 length byte array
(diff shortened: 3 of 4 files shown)


rhatdan:
This is a breaking change, but I think it is worth it, and probably affects almost nobody.

rhatdan:
@nalind PTAL

rhatdan:
@thaJeztah @kolyshkin @mrunalp PTAL

go-selinux/selinux.go
```diff
@@ -62,15 +62,28 @@ func ClassIndex(class string) (int, error) {
 }
 
 // SetFileLabel sets the SELinux label for this path or returns an error.
+// follow symlinks
```
kolyshkin, line 65: This have to be a complete sentence.

go-selinux/selinux.go
```diff
@@ -62,15 +62,28 @@ func ClassIndex(class string) (int, error) {
...
 func SetFileLabel(fpath string, label string) error {
 	return setFileLabel(fpath, label)
 }
 
+// SetFileLabel sets the SELinux label for this path or returns an error.
```
kolyshkin, line 70: `s/SetFileLabel/LSetFileLabel/g`

go-selinux/selinux.go
```diff
@@ -62,15 +62,28 @@ func ClassIndex(class string) (int, error) {
...
+// does not follow symlinks
```
kolyshkin, line 71: _If the argument is a symbolic link, the label is set on the link itself. This function makes no attempt to follow the symlink._

Or somesuch.

go-selinux/selinux.go
```diff
@@ -62,15 +62,28 @@ func ClassIndex(class string) (int, error) {
...
+	return lsetFileLabel(fpath, label)
+}
+
 // FileLabel returns the SELinux label for this path or returns an error.
 func FileLabel(fpath string) (string, error) {
 	return fileLabel(fpath)
 }
 
+// LFileLabel returns the SELinux label for this path does not follow symlinks
+// or returns an error.
```
kolyshkin, line 82: Does not look like a valid English sentence.

go-selinux/selinux_linux.go
```diff
@@ -317,7 +317,7 @@ func classIndex(class string) (int, error) {
 }
 
 // setFileLabel sets the SELinux label for this path or returns an error.
```
kolyshkin, line 319: `s/setFileLabel/lsetFileLabel/`

go-selinux/selinux.go
```diff
@@ -62,15 +62,28 @@ func ClassIndex(class string) (int, error) {
...
 
 // SetFileLabel sets the SELinux label for this path or returns an error.
+// follow symlinks
 func SetFileLabel(fpath string, label string) error {
 	return setFileLabel(fpath, label)
 }
 
+// SetFileLabel sets the SELinux label for this path or returns an error.
+// does not follow symlinks
+func LSetFileLabel(fpath string, label string) error {
```
kolyshkin, line 72: Maybe it should be named `LsetFileLabel` as other similar functions have lowercase letter after L (os.Lstat, unix.Lsetxattr).

go-selinux/selinux_linux.go
```diff
@@ -334,12 +334,47 @@ func setFileLabel(fpath string, label string) error {
...
+		return "", &os.PathError{Op: "lgetxattr", Path: fpath, Err: err}
+	}
+	// Trim the NUL byte at the end of the byte buffer, if present.
+	if len(label) > 0 && label[len(label)-1] == '\x00' {
+		label = label[:len(label)-1]
+	}
+	return string(label), nil
+}
+
+// fileLabel returns the SELinux label for this path or returns an error.
```
kolyshkin, line 372: function name is wrong

go-selinux/selinux_linux.go
```diff
@@ -334,12 +334,47 @@ func setFileLabel(fpath string, label string) error {
...
+func lfileLabel(fpath string) (string, error) {
+	if fpath == "" {
+		return "", ErrEmptyPath
+	}
```
kolyshkin, line 376: I would use `lFileLabel`.

go-selinux/selinux_linux.go
```diff
@@ -334,12 +334,47 @@ func setFileLabel(fpath string, label string) error {
...
+
 // fileLabel returns the SELinux label for this path or returns an error.
 func fileLabel(fpath string) (string, error) {
 	if fpath == "" {
 		return "", ErrEmptyPath
 	}
 
+	label, err := getxattr(fpath, xattrNameSelinux)
+	if err != nil {
+		return "", &os.PathError{Op: "lgetxattr", Path: fpath, Err: err}
```
kolyshkin, line 363: typo: `/slgetxattr/getxattr/`

go-selinux/xattrs_linux.go
```diff
@@ -36,3 +36,36 @@ func doLgetxattr(path, attr string, dest []byte) (int, error) {
...
+	sz, errno := dogetxattr(path, attr, dest)
+	for errno == unix.ERANGE { //nolint:errorlint // unix errors are bare
+		// Buffer too small, use zero-sized buffer to get the actual size
+		sz, errno = dogetxattr(path, attr, []byte{})
+		if errno != nil {
+			return nil, errno
+		}
+
+		dest = make([]byte, sz)
+		sz, errno = doLgetxattr(path, attr, dest)
```
kolyshkin, line 54: Typo? `s/doLgetxattr/dogetxattr/`

kolyshkin: @rhatdan ^^^

rhatdan:
@kolyshkin @thaJeztah I think i have fixed all my cut and paste errors. PTAL

go-selinux/selinux_linux.go
```diff
@@ -316,8 +316,8 @@ func classIndex(class string) (int, error) {
 	return index, nil
 }
 
-// setFileLabel sets the SELinux label for this path or returns an error.
-func setFileLabel(fpath string, label string) error {
+// lSetFileLabel sets the SELinux label for this path or returns an error.
```
thaJeztah, line 319: nit: I guess this one also could use the "not following symlinks"

go-selinux/selinux_linux.go
```diff
@@ -334,12 +334,47 @@ func setFileLabel(fpath string, label string) error {
 	return nil
 }
 
+// setFileLabel sets the SELinux label for this path or returns an error.
```
thaJeztah, line 337: nit: same

go-selinux/selinux_linux.go
```diff
@@ -334,12 +334,47 @@ func setFileLabel(fpath string, label string) error {
...
+		return "", &os.PathError{Op: "getxattr", Path: fpath, Err: err}
+	}
+	// Trim the NUL byte at the end of the byte buffer, if present.
+	if len(label) > 0 && label[len(label)-1] == '\x00' {
+		label = label[:len(label)-1]
+	}
+	return string(label), nil
+}
+
+// lFileLabel returns the SELinux label for this path or returns an error.
```
thaJeztah, line 372: nit: same

thaJeztah (approved):
code LGTM

the remaining nits are for non-exported functions, not a show-stopper per-se

thaJeztah:
The only thing I'm wondering (w.r.t. breaking change); could there be cases where labels are set on `/var/run/....` (which usually is a symlink to `/run/...` ?

rhatdan:
I think this change would not effect a path with a symlink in it, it only effects if the basename is a symlink.
So if you were trying to label /var/run, you would now label /run, and leave the /var/run symlink alone.

But if you were labeling /var/run/foobar, and foobar was a file, it would label the file correctly.

kolyshkin (approved):
LGTM

thaJeztah:
> I think this change would not effect a path with a symlink in it, it only effects if the basename is a symlink. So if you were trying to label /var/run, you would now label /run, and leave the /var/run symlink alone.
> 
> But if you were labeling /var/run/foobar, and foobar was a file, it would label the file correctly.

Thanks! Makes sense (should probably have looked closer); agreed I don't expect that to be too problematic.

(post-merge LGTM - saw you addressed those last nits; thanks!)

PR #177, opened 2022-06-30
If chcon fails, check if label is already correct

rhatdan:
Currently if a user attempts to chcon a file or directory and fails for
any reason check if the file already has the right label, and continue.

### DIFF
--- a/go-selinux/rchcon.go
+++ b/go-selinux/rchcon.go
@@ -12,7 +12,18 @@ import (
 )
 
 func rchcon(fpath, label string) error {
+	fastMode := false
+	// If the current label matches the new label, assume
+	// other labels are correct.
+	if cLabel, err := lFileLabel(fpath); err == nil && cLabel == label {
+		fastMode = true
+	}
 	return pwalkdir.Walk(fpath, func(p string, _ fs.DirEntry, _ error) error {
+		if fastMode {
+			if cLabel, err := lFileLabel(fpath); err == nil && cLabel == label {
+				return nil
+			}
+		}
 		e := lSetFileLabel(p, label)
 		// Walk a file tree can race with removal, so ignore ENOENT.
 		if errors.Is(e, os.ErrNotExist) {
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -1084,7 +1084,24 @@ func chcon(fpath string, label string, recurse bool) error {
 	}
 
 	if !recurse {
-		return setFileLabel(fpath, label)
+		err := lSetFileLabel(fpath, label)
+		if err != nil {
+			// Check if file doesn't exist, must have been removed
+			if errors.Is(err, os.ErrNotExist) {
+				return nil
+			}
+			// Check if current label is correct on disk
+			flabel, nerr := lFileLabel(fpath)
+			if nerr == nil && flabel == label {
+				return nil
+			}
+			// Check if file doesn't exist, must have been removed
+			if errors.Is(nerr, os.ErrNotExist) {
+				return nil
+			}
+			return err
+		}
+		return nil
 	}
 
 	return rchcon(fpath, label)


rhatdan:
@giuseppe @thaJeztah @kolyshkin PTAL

rhatdan:
Attempt to help fix: https://github.com/containers/podman/issues/14786

rhatdan:
[jkroon81](https://github.com/jkroon81) PTAL

giuseppe (approved):
LGTM

thaJeztah:
I think golangci-lint must be updated to a newer version to work with Go 1.18 (we can likely drop older go versions from the CI matrix as well)

rhatdan:
@thaJeztah How do I update lint?

go-selinux/rchcon.go
```diff
@@ -12,7 +12,22 @@ import (
...
 
 func rchcon(fpath, label string) error {
+	slowMode := false
+	// If the current label matches the new label, assume
+	// other labels are correct.
+	if currentLabel, err := lFileLabel(fpath); err == nil {
+		if label == currentLabel {
+			slowMode = true
+		}
+	}
```
thaJeztah, line 22: Slightly wondering if it's easier to read if it's a one-liner;

```suggestion
	if currentLabel, err := lFileLabel(fpath); err == nil && label == currentLabel {
		slowMode = true
	}
```

go.mod
```diff
@@ -1,5 +1,5 @@
 module github.com/opencontainers/selinux
 
-go 1.13
+go 1.18
 
-require golang.org/x/sys v0.0.0-20191115151921-52ab43148777
+require golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a
```
thaJeztah, line 5: Might be good to do this in a separate commit (keeping the "fix" separate from assisting changes)

go-selinux/rchcon.go
```diff
@@ -12,7 +12,22 @@ import (
 )
 
 func rchcon(fpath, label string) error {
+	slowMode := false
```
thaJeztah, line 15: Perhaps a silly question, but what does "slowMode" mean here? Looking at the changes, it seems that `slowMode` actually enables a "fast path" (check the label, and we're done), but the name of this variable feels like it has the reverse meaning 😅  

Or is reading the label (`lFileLabel()`) slower than setting it (` lSetFileLabel()`) ?

go-selinux/rchcon.go
thaJeztah, line 22: Slightly wondering if it'd be easier to read if this was a one-liner;

```suggestion
	if currentLabel, err := lFileLabel(fpath); err == nil && label == currentLabel {
		slowMode = true
	}
```

go-selinux/selinux_linux.go
```diff
@@ -1102,7 +1102,18 @@ func chcon(fpath string, label string, recurse bool) error {
...
-		return setFileLabel(fpath, label)
+		err := lSetFileLabel(fpath, label)
+		if err == nil {
+			return nil
+		}
+		if errors.Is(err, os.ErrNotExist) {
+			return err
+		}
+		flabel, _ := lFileLabel(fpath)
+		if flabel == label {
```
thaJeztah, line 1113: I _guess_ it's ok(isn), but we don't check the error here, so technically can't assume `flabel` is an ok value

go-selinux/selinux_linux.go
```diff
@@ -1102,7 +1102,18 @@ func chcon(fpath string, label string, recurse bool) error {
...
+			return nil
+		}
+		return err
```
thaJeztah, line 1116: Somewhat related to my earlier question; if reading back the label is faster than setting it, should we reverse this?

Currently, this;

- tries to set the label
- if that fails for _whatever reason_, other than "file doesn't exist", we try to read the existing label
- if that fails, we return the first error

If reading is indeed faster, we could do;


```suggestion
		if flabel, err := lFileLabel(fpath); err == nil && flabel == label {
			// already has the right label
			return nil
		}

		return lSetFileLabel(fpath, label)
```

rhatdan:
@thaJeztah PTAL again.

rhatdan:
@mrunalp @vrothberg  @kolyshkin PTAL

PR #180, opened 2022-09-23
Allow relabeling of files/directories under /usr prefix

rhatdan:
We already check to make sure users do not accidentily relabel these excluded paths:

	exclude_paths := map[string]bool{
		"/":           true,
		"/bin":        true,
		"/boot":       true,
		"/dev":        true,
		"/etc":        true,
		"/etc/passwd": true,
		"/etc/pki":    true,
		"/etc/shadow": true,
		"/home":       true,
		"/lib":        true,
		"/lib64":      true,
		"/media":      true,
		"/opt":        true,
		"/proc":       true,
		"/root":       true,
		"/run":        true,
		"/sbin":       true,
		"/srv":        true,
		"/sys":        true,
		"/tmp":        true,
		"/usr":        true,
		"/var":        true,
		"/var/lib":    true,
		"/var/log":    true,
	}

But some users put homedirectories under /usr, and I see no reason to block them from relabeling.

At a certain point if users do something dumb with relableing we can not stop them.

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -1072,21 +1072,6 @@ func copyLevel(src, dest string) (string, error) {
 	return tcon.Get(), nil
 }
 
-// Prevent users from relabeling system files
-func badPrefix(fpath string) error {
-	if fpath == "" {
-		return ErrEmptyPath
-	}
-
-	badPrefixes := []string{"/usr"}
-	for _, prefix := range badPrefixes {
-		if strings.HasPrefix(fpath, prefix) {
-			return fmt.Errorf("relabeling content in %s is not allowed", prefix)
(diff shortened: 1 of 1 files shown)


rhatdan:
@kolyshkin @thaJeztah @vrothberg @giuseppe @mrunalp PTAL

rhatdan:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2124430

giuseppe (approved):
LGTM

vrothberg (approved):
LGTM

rhatdan:
We could move it into that function instead of label.Relabel.  Where we use this in Podman and Buildah is always throught lable.Relabel.

PR #181, opened 2022-09-26
selinux.Chcon should check legal rather then just label.Relabel

rhatdan:
Since label.Relabel ends up calling into selinux.chcon, we should do the check for invalid directories under chcon. This will allow the selinux.Chcon function to also be verified.


rhatdan:
@thaJeztah @mrunalp @vrothberg @giuseppe @kolyshkin PTAL

go-selinux/selinux_linux.go
```diff
@@ -1083,6 +1084,50 @@ func chcon(fpath string, label string, recurse bool) error {
...
+
+	if home := os.Getenv("HOME"); home != "" {
+		exclude_paths[home] = true
+	}
+
+	if sudoUser := os.Getenv("SUDO_USER"); sudoUser != "" {
+		if usr, err := user.Lookup(sudoUser); err == nil {
+			exclude_paths[usr.HomeDir] = true
+		}
+	}
```
thaJeztah, line 1122: Will this part bring back the issue you ran into that you were trying to solve in https://github.com/opencontainers/selinux/pull/180?

> But some users put homedirectories under /usr, and I see no reason to block them from relabeling.

rhatdan: No the #180 was blocking /usr/*
Now we are just blocking /usr or $HOME not $HOME/*

thaJeztah: Ah, gotcha!

thaJeztah (approved):
LGTM, thanks!

rhatdan:
@vrothberg @giuseppe @kolyshkin @mrunalp PTAL

vrothberg (commented):
LGTM

PR #183, opened 2022-10-05
gha: add go1.19, drop go1.16, update golangci-lint to v1.49

thaJeztah:
See individual commits for details

PR #185, opened 2022-10-05
pkg/pwalk: deprecate in favor of pkg/pwalkdir

thaJeztah:
The package was already deprecated in the README, but not in code. This patch marks the code as deprecated, as go1.15 reached EOL, so all consumers should be able to use the new package.

In a follow-up, we can make these functions an alias for the new package (or remove it altogether).

### DIFF
--- a/pkg/pwalk/pwalk.go
+++ b/pkg/pwalk/pwalk.go
@@ -8,6 +8,10 @@ import (
 	"sync"
 )
 
+// WalkFunc is the type of the function called by Walk to visit each
+// file or directory. It is an alias for [filepath.WalkFunc].
+//
+// Deprecated: use [github.com/opencontainers/selinux/pkg/pwalkdir] and [fs.WalkDirFunc].
 type WalkFunc = filepath.WalkFunc
 
 // Walk is a wrapper for filepath.Walk which can call multiple walkFn
@@ -29,6 +33,8 @@ type WalkFunc = filepath.WalkFunc
 // - if more than one walkFn instance will return an error, only one
 // of such errors will be propagated and returned by Walk, others
 // will be silently discarded.
+//
+// Deprecated: use [github.com/opencontainers/selinux/pkg/pwalkdir.Walk]
 func Walk(root string, walkFn WalkFunc) error {
 	return WalkN(root, walkFn, runtime.NumCPU()*2)
 }
@@ -38,6 +44,8 @@ func Walk(root string, walkFn WalkFunc) error {
 // num walkFn will be called at any one time.
 //
 // Please see Walk documentation for caveats of using this function.
+//
+// Deprecated: use [github.com/opencontainers/selinux/pkg/pwalkdir.WalkN]
 func WalkN(root string, walkFn WalkFunc, num int) error {
 	// make sure limit is sensible
 	if num < 1 {


thaJeztah:
@kolyshkin @mrunalp @rhatdan PTAL

pkg/pwalk/pwalk.go
```diff
@@ -29,16 +29,24 @@ type WalkFunc = filepath.WalkFunc
 // - if more than one walkFn instance will return an error, only one
 // of such errors will be propagated and returned by Walk, others
 // will be silently discarded.
+//
+// Deprecated: use github.com/opencontainers/selinux/pwalkdir.Walk
```
kolyshkin, line 33: Do you mean `github.com/opencontainers/selinux/pkg/pwalkdir.Walk`

pkg/pwalk/pwalk.go
```diff
@@ -29,16 +29,24 @@ type WalkFunc = filepath.WalkFunc
...
 // num walkFn will be called at any one time.
 //
 // Please see Walk documentation for caveats of using this function.
+//
+// Deprecated: use github.com/opencontainers/selinux/pwalkdir.WalkN
 func WalkN(root string, walkFn WalkFunc, num int) error {
+	return walkN(root, walkFn, num)
+}
+
+func walkN(root string, walkFn WalkFunc, num int) error {
```
kolyshkin, line 49: What is the meaning of this change? Is that for one deprecated function to not use another? In this case perhaps it's easier to annotate instead.

pkg/pwalk/pwalk.go
thaJeztah: Ah, yes,  missed the `pkg/`

pkg/pwalk/pwalk.go
thaJeztah: Yeah, both my IDE and linter were complaining about using deprecated functions; so could've either added a `//nolint:<xxx>` flag, or do this.


I'm wondering though if we perhaps should immediately proceed with aliasing this to the new package, if we're going to support go1.16+ (https://github.com/opencontainers/selinux/pull/183), WDYT?

thaJeztah: I changed this back.

w.r.t. aliasing; we can't alias it as the signatures don't match, so let's deprecate it as step 1, then remove after

thaJeztah:
@kolyshkin updated; PTAL

rhatdan:
LGTM

thaJeztah:
@rhatdan you need to use GitHub's "approve" now (we're no longer using pullapprove on this repo 😁)

rhatdan (approved):

rhatdan:
@kolyshkin @giuseppe PTAL

pkg/pwalk/pwalk.go
```diff
@@ -29,6 +33,8 @@ type WalkFunc = filepath.WalkFunc
 // - if more than one walkFn instance will return an error, only one
 // of such errors will be propagated and returned by Walk, others
 // will be silently discarded.
+//
+// Deprecated: use [github.com/opencontainers/selinux/pkg/pwalkdir.Walk]
```
kolyshkin, line 37: Kudos for using doc links!

Nit: missing period at EOL.

pkg/pwalk/pwalk.go
```diff
@@ -38,6 +44,8 @@ func Walk(root string, walkFn WalkFunc) error {
 // num walkFn will be called at any one time.
 //
 // Please see Walk documentation for caveats of using this function.
+//
+// Deprecated: use [github.com/opencontainers/selinux/pkg/pwalkdir.WalkN]
```
kolyshkin, line 48: Nit: missing period at EOL.

pkg/pwalk/pwalk.go
thaJeztah: Let me try it again locally, but I think there may have been a bug in Golang's parser; the doc links only work when surrounded by whitespace (to prevent `map[string]string` being seen as a link), but that also meant that adding a period at the end cause it not to be seen as a link 😞 

(Haven't filed an issue for it yet, but did for another one; https://github.com/golang/go/issues/56072)

PR #186, opened 2022-10-05
remove vendor directory

thaJeztah:
- follow-up to / depends on https://github.com/opencontainers/selinux/pull/183

Using a vendor directory is common for repositories that produce binaries,
but usually not for library-only repositories. Having a vendor directory
can be useful to review changes in dependencies, but as this repository
only has a single dependency (which is well-known and (usually) requires
minimal review), it's probably ok to drop the vendor directory.


rhatdan:
LGTM
@kolyshkin @giuseppe @vrothberg @mrunalp PTAL

rhatdan:
@lsm5 PTAL

giuseppe (approved):
LGTM

thaJeztah:
We probably need https://github.com/opencontainers/selinux/pull/183 first (this one includes those changes, but I can  rebase it after that).

thaJeztah:
rebased; this should be ready for review now, PTAL 👍

PR #187, opened 2022-10-06
go-selinux: remove redundant stub for DisableSecOpt()

thaJeztah:
Both implementations were the same, so no stub is needed.


rhatdan:
LGTM

PR #188, opened 2022-10-10
use filepath instead of path

thaJeztah:
Originally, this package was documented to be for paths separated by forward slashes, but I was recently pointed to the updated package description, which amended the documentation (see https://go-review.googlesource.com/32423) and added:

> To manipulate operating system paths, use the path/filepath package.

A later patch (https://go-review.googlesource.com/45653) tweaked the description again, adding to the `path/filepath` package:

> To process paths such as URLs that always use forward slashes regardless of
> the operating system, see the path package.

And to the `path` package

> This package does not deal with Windows paths with drive letters or backslashes;
> to manipulate operating system paths, use the path/filepath package.

I expect it to be "fine" in general to use the `path` package for Unix/Linux, but
from both patches, the intent of the Go maintainers is to use `path/filepath`
for anything file(system) related, so let's adapt to that.

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -11,7 +11,6 @@ import (
 	"math/big"
 	"os"
 	"os/user"
-	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@kolyshkin @giuseppe @lsm5 PTAL

kolyshkin (approved):
I am slightly more inclined towards using `path` rather than `filepath`, as `path` is somewhat simpler, and will definitely work fine on Linux (and this code is Linux-specific and will stay that way).

OTOH I don't see a reason to not use filepath, so LGTM.

thaJeztah:
Yup; I originally thought "path" was the way to go as well. Fun fact (but I need to give it another  go), was that I did some benchmarking at some time, and `filepath` was more performant than `path` (perhaps they added some optimisation that they didn't add to `path` ?

Also if their _intent_ is for `filepath` to be used for file paths, then there is "some" risk that `path` may be missing some (obscure?) handling for file-specifics, or at least their focus may not be on that, so using `filepath` would be the safer option I guess?

PR #190, opened 2022-10-10
go-selinux: some cleanup/refactor and improve test coverage

thaJeztah:
See individual commits for details

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -390,21 +390,19 @@ func lFileLabel(fpath string) (string, error) {
 	return string(label), nil
 }
 
-// setFSCreateLabel tells kernel the label to create all file system objects
-// created by this task. Setting label="" to return to default.
 func setFSCreateLabel(label string) error {
-	return writeAttr("fscreate", label)
+	return writeCon(attrPath("fscreate"), label)
 }
 
 // fsCreateLabel returns the default label the kernel which the kernel is using
 // for file system objects created by this task. "" indicates default.
 func fsCreateLabel() (string, error) {
-	return readAttr("fscreate")
+	return readCon(attrPath("fscreate"))
 }
 
 // currentLabel returns the SELinux label of the current process thread, or an error.
 func currentLabel() (string, error) {
-	return readAttr("current")
+	return readCon(attrPath("current"))
 }
 
 // pidLabel returns the SELinux label of the given pid, or an error.
@@ -415,7 +413,7 @@ func pidLabel(pid int) (string, error) {
 // ExecLabel returns the SELinux label that the kernel will use for any programs
 // that are executed by the current process thread, or an error.
 func execLabel() (string, error) {
-	return readAttr("exec")
+	return readCon(attrPath("exec"))
 }
 
 func writeCon(fpath, val string) error {
@@ -465,14 +463,6 @@ func attrPath(attr string) string {
 	}
 
 	return path.Join("/proc/self/task/", strconv.Itoa(unix.Gettid()), "/attr/", attr)
-}
-
-func readAttr(attr string) (string, error) {
-	return readCon(attrPath(attr))
-}
-
-func writeAttr(attr, val string) error {
-	return writeCon(attrPath(attr), val)
 }
 
 // canonicalizeContext takes a context string and writes it to the kernel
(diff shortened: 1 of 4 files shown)


go-selinux/selinux_linux_test.go
```diff
@@ -25,7 +25,8 @@ func TestSetFileLabel(t *testing.T) {
 	out.Close()
 	defer os.Remove(tmp)
 
-	if err := SetFileLabel(tmp, con); err != nil {
+	err = SetFileLabel(tmp, con)
+	if err != nil {
```
kolyshkin, line 29: I don't understand the reason for this change.

go-selinux/selinux_linux_test.go
```diff
@@ -514,7 +515,8 @@ func BenchmarkChcon(b *testing.B) {
 	b.Logf("Chcon(%q, %q)", dir, con)
 	b.ResetTimer()
 	for n := 0; n < b.N; n++ {
-		if err := Chcon(dir, con, true); err != nil {
+		err = Chcon(dir, con, true)
+		if err != nil {
```
kolyshkin, line 519: ditto

go-selinux/selinux_linux.go
```diff
@@ -549,30 +549,30 @@ func (l *level) parseLevel(levelStr string) error {
 
 // rangeStrToMLSRange marshals a string representation of a range.
 func rangeStrToMLSRange(rangeStr string) (*mlsRange, error) {
-	mlsRange := &mlsRange{}
+	mlsrange := &mlsRange{}
```
kolyshkin, line 552: can we just use `range` (or even `r`) here?

go-selinux/selinux_linux.go
```diff
@@ -712,11 +712,11 @@ func readWriteCon(fpath string, val string) (string, error) {
 
 // peerLabel retrieves the label of the client on the other side of a socket
 func peerLabel(fd uintptr) (string, error) {
-	label, err := unix.GetsockoptString(int(fd), unix.SOL_SOCKET, unix.SO_PEERSEC)
+	lbl, err := unix.GetsockoptString(int(fd), unix.SOL_SOCKET, unix.SO_PEERSEC)
```
kolyshkin, line 715: perhaps just `l`?

go-selinux/selinux_linux.go
```diff
@@ -734,8 +734,8 @@ func setKeyLabel(label string) error {
 
 // get returns the Context as a string
 func (c Context) get() string {
-	if level := c["level"]; level != "" {
-		return c["user"] + ":" + c["role"] + ":" + c["type"] + ":" + level
+	if lvl := c["level"]; lvl != "" {
+		return c["user"] + ":" + c["role"] + ":" + c["type"] + ":" + lvl
```
kolyshkin, line 738: perhaps just `l`?

go-selinux/selinux_linux_test.go
thaJeztah: not a huge one; my editor's linter warns about variables that are shadowed; in this case an `err` already existed. I could just remove the `:`, but thought it's more clear to break it in two lines if we're reusing the variable.

go-selinux/selinux_linux.go
thaJeztah: `range` is a reserved keyboard. Can use `r` if we think it's clear enough

go-selinux/selinux_linux.go
```diff
@@ -606,17 +606,17 @@ func bitsetToStr(c *big.Int) string {
 	return str
 }
 
-func (l1 *level) equal(l2 *level) bool {
```
kolyshkin, line 619: I like `l1` and `l2` more in this case, since we compare them. A linter is not always right.

kolyshkin (changes requested):
Commits 1 and 4 LGTM, left a few comments for 2 and 3.

go-selinux/selinux_linux_test.go
kolyshkin: So, in this case the linter warning should be ignored. Yes, we have a variable named `err` already, and yes, here we create another `err` which shadows the original one. No, there is no issue here.

Or maybe just remove the `:`.

kolyshkin:
One thing that bothers me if those linter warnings you fix here will resurface after some time, as apparently we don't have those linters enabled in CI. So, if there is a linter in golangci-lint that warns of these, the PR would make more sense.

thaJeztah:
@kolyshkin updated; moved some of the changes to https://github.com/opencontainers/selinux/pull/191, which enables more linters (some of which catching the shadowing issues, although I haven't enabled "strict" mode, which seems to closer match what my IDE shows)

rhatdan:
LGTM

PR #191, opened 2022-10-12
golangci-lint: enable more linters and address linting issues

thaJeztah:
- follow-up to https://github.com/opencontainers/selinux/pull/189

(more coming after this, but I have ETOOMANYBRANCHES already)

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -36,7 +36,7 @@ jobs:
           go-version: 1.19.x
       - uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.49
+          version: v1.50
 
   cross:
     runs-on: ubuntu-20.04
@@ -54,7 +54,7 @@ jobs:
           go-version: 1.19.x
       - uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.49
+          version: v1.50
       - name: test-stubs
         run: make test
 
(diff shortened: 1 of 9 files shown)


rhatdan:
I am ok with almost all of these, although I am not crazy about the `err:=` shadow issue.  I find it a lot easier to read

if err := foobar(); err != nil 
}

Then 
err := foobar()
if err != nil {
}

Bottom line, I think err should be handled differently then other shadowed vars.  But I am willing to live with the changes.

LGTM

thaJeztah:
>  I am not crazy about the `err:=` shadow issue.

Yes, I'm slightly on the fence on that one as well;

- in _most_ cases, it's mostly "noise"
- ^^ that said, I'm not against doing a "once in a while" round on changing them
- ^^^ but ... that said; I've also encountered situations where an `err` was masking an `err` that was checked in a `defer()`. Admitted, those situations are rare (and usually better to have a distinct name for "special" errors).

I don't think the linter itself has a config for this, but perhaps a string match exclusion would work (see https://golangci-lint.run/usage/false-positives/#exclude-issue-by-text)

rhatdan:
Tests are failing for go-fumpt?

thaJeztah:
> Tests are failing for go-fumpt?

I probably need to rebase see https://github.com/opencontainers/selinux/pull/189#issuecomment-1277485676

thaJeztah:
Rebased, and tweaked the settings to ignore "err" for shadowing 👍  PTAL

thaJeztah:
Oh! Had one more change staged in another branch; just pushed that one as well.

rhatdan:
Nice work @thaJeztah

vrothberg (commented):
LGTM

PR #192, opened 2022-10-17
Adds MLSEnabled function.

### LINKED ISSUE #176: Checking if MLS is enabled
Is there a way to check using this library if MLS is enabled on a system?

badochov:
As in https://github.com/SELinuxProject/selinux/blob/master/libselinux/src/enabled.c#L29

Closes #176.

### DIFF
--- a/go-selinux/selinux.go
+++ b/go-selinux/selinux.go
@@ -213,6 +213,11 @@ func ReserveLabel(label string) {
 	reserveLabel(label)
 }
 
+// MLSEnabled checks if MLS is enabled.
+func MLSEnabled() bool {
+	return isMLSEnabled()
+}
+
 // EnforceMode returns the current SELinux mode Enforcing, Permissive, Disabled
 func EnforceMode() int {
 	return enforceMode()
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -781,6 +781,15 @@ func selinuxEnforcePath() string {
 	return filepath.Join(getSelinuxMountPoint(), "enforce")
 }
 
+// isMLSEnabled checks if MLS is enabled.
+func isMLSEnabled() bool {
+	enabledB, err := os.ReadFile(filepath.Join(getSelinuxMountPoint(), "mls"))
+	if err != nil {
+		return false
+	}
+	return bytes.Equal(enabledB, []byte{'1'})
+}
+
 // enforceMode returns the current SELinux mode Enforcing, Permissive, Disabled
 func enforceMode() int {
 	var enforce int
--- a/go-selinux/selinux_stub.go
+++ b/go-selinux/selinux_stub.go
@@ -95,6 +95,10 @@ func clearLabels() {
 func reserveLabel(label string) {
 }
 
+func isMLSEnabled() bool {
+	return false
+}
+
 func enforceMode() int {
 	return Disabled
 }


rhatdan:
LGTM
@thaJeztah @kolyshkin @vrothberg @giuseppe PTAL

go-selinux/selinux_linux.go
```diff
@@ -781,6 +781,20 @@ func selinuxEnforcePath() string {
 	return filepath.Join(getSelinuxMountPoint(), "enforce")
 }
 
+func selinuxMLSEnabledPath() string {
+	return filepath.Join(getSelinuxMountPoint(), "mls")
+}
+
+// isMLSEnabled checks if MLS is enabled.
+func isMLSEnabled() bool {
+	enabledB, err := os.ReadFile(selinuxMLSEnabledPath())
```
thaJeztah, line 790: I see it's done in other places as well, but I think the `selinuxMLSEnabledPath()` may be a bit "too much abstraction", especially as  we only use it in a single place; perhaps just inline it;

```go
	enabledB, err := os.ReadFile(filepath.Join(getSelinuxMountPoint(), "mls"))
```

go-selinux/selinux_linux.go
```diff
@@ -781,6 +781,20 @@ func selinuxEnforcePath() string {
...
+	if err != nil {
+		return false
+	}
+
+	return !bytes.Equal(enabledB, []byte{'1'})
```
thaJeztah, line 795: Should this be the reverse? (i.e., it return `true` if the file contains `1` ?

go-selinux/selinux.go
```diff
@@ -213,6 +213,11 @@ func ReserveLabel(label string) {
 	reserveLabel(label)
 }
 
+// IsMLSEnabled checks if MLS is enabled.
+func IsMLSEnabled() bool {
+	return isMLSEnabled()
```
thaJeztah, line 218: I think we're missing a "stub" implementation for `isMLSEnabled()` in `selinux_unsupported.go`; 

Screenshot 2022-10-19 at 13 51 48


(This may also explain the other Linter issues?)

go-selinux/selinux.go
```diff
@@ -213,6 +213,11 @@ func ReserveLabel(label string) {
 	reserveLabel(label)
 }
 
+// IsMLSEnabled checks if MLS is enabled.
+func IsMLSEnabled() bool {
```
thaJeztah, line 217: It's more conventional in Go to omit `Is` prefixes (as the `bool` return already describes it's a boolean result). (Same for `Get` (and `Set`), but that's already in the code, so we can't reverse that 😂).

Perhaps (same for the non-exported variants);

```go
// MLSEnabled checks if MLS is enabled.
func MLSEnabled() bool {
```

Which will be called as `selinux.MLSEnabled()`, e.g.

```go
if selinux.MLSEnabled() {
    // do magic stuff!
}
```

thaJeztah (changes requested):
Thanks! left comments inline

go-selinux/selinux_linux.go
badochov: It is weird but apparently not https://github.com/SELinuxProject/selinux/blob/3f11c7d19c449996d7c34190d6aa74220b104779/libselinux/src/enabled.c#L51

go-selinux/selinux.go
badochov: right

go-selinux/selinux_linux.go
thaJeztah: Ooooh!

```
/*
 * Function: is_selinux_mls_enabled()
 * Return:   1 on success
 *	     0 on failure
 */
```

Hm.. right; perhaps would be good to add a comment there to describe it (perhaps also with that link) to explain this is really correct 😅.

go-selinux/selinux_stub.go
```diff
@@ -95,6 +95,9 @@ func clearLabels() {
 func reserveLabel(label string) {
 }
 
+func isMLSEnabled() bool {
+}
```
thaJeztah, line 99: This needs a return;

```go
func isMLSEnabled() bool {
	return false
}
```

thaJeztah:
@badochov you may have missed my last comment;

> This needs a return;
> 
> ```go
> func isMLSEnabled() bool {
> 	return false
> }
> ```


While updating, could you also update the commit message (change `IsMLSEnabled` to `MLSEnabled `)?

thaJeztah:
Oh, never mind, I see you also didn't address the other comment yet 😅 (let me know when it's updated 👍 )

badochov:
@thaJeztah I've updated the PR.

thaJeztah (approved):
LGTM, thanks!

go-selinux/selinux_linux.go
badochov: @giuseppe yes you're right

badochov: i was going through this line and documentation of strcmo multiple times I have no idea how I convoced myself to be wrong so many times

thaJeztah: OH MY! 🤦 for some reason I completely didn't spot that as well. Thanks @giuseppe !!! (it adds to my dislike of "too many booleans")

badochov: fixed

thaJeztah (approved):
LGTM, thanks for fixing!

@giuseppe PTAL

giuseppe (approved):
Thanks

LGTM

thaJeztah:
Let me bring this one in; I see @rhatdan already gave a "conceptual" approval, so looks like it should be good to go.

Thanks @badochov !

PR #193, opened 2022-11-01
feat: dependabot workflow automation for updating dependency

Rajpratik71:
Signed-off-by: Pratik Raj <Rajpratik71@gmail.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  # Dependencies listed in go.mod
+  - package-ecosystem: "gomod"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+
+  # Dependencies listed in .github/workflows/*.yml
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


rhatdan:
LGTM

thaJeztah:
I'm a bit on the fence on enabling dependabot, for a couple of reasons;

- go modules specify the _minimum_ required version to use the module, leaving it to consumers of the module to use newer versions (if there's a need in those). Updating the dependency for no particular reason means that all consumers of this repository are _also_ forced to update the dependency (which can be rather disruptive), and goes against the principle of go modules specifying the _minimum_ required version.
- This repository only has a single non-core dependency (golang.org/x/sys) and the only reason to update the dependency would be if there's either a severe bug, or a security issue in that dependency in the code we use (which is very small subset of the golang.org/x/sys module, see below)
- dependabot opens pull requests by creating branches on the upstream (this) repository (not a huge fan of that)
- dependabot _cannot be disabled_ on forks. With golang.org/x/sys being a large module with many updates, this means that it will open a pull request _every week_ on _all_ forks, causing a lot of noise.


A quick look at what parts we use of the dependency;

```
git grep 'unix\.'
go-selinux/selinux_linux.go:    var buf unix.Statfs_t
go-selinux/selinux_linux.go:            err := unix.Statfs(mnt, &buf)
go-selinux/selinux_linux.go:            if err == unix.EAGAIN || err == unix.EINTR { //nolint:errorlint // unix errors are bare
go-selinux/selinux_linux.go:    if uint32(buf.Type) != uint32(unix.SELINUX_MAGIC) {
go-selinux/selinux_linux.go:    if (buf.Flags & unix.ST_RDONLY) != 0 {
go-selinux/selinux_linux.go:    var buf unix.Statfs_t
go-selinux/selinux_linux.go:            err := unix.Fstatfs(int(fh.Fd()), &buf)
go-selinux/selinux_linux.go:            if err != unix.EINTR { //nolint:errorlint // unix errors are bare
go-selinux/selinux_linux.go:    if buf.Type != unix.PROC_SUPER_MAGIC {
go-selinux/selinux_linux.go:            err := unix.Lsetxattr(fpath, xattrNameSelinux, []byte(label), 0)
go-selinux/selinux_linux.go:            if err != unix.EINTR { //nolint:errorlint // unix errors are bare
go-selinux/selinux_linux.go:            err := unix.Setxattr(fpath, xattrNameSelinux, []byte(label), 0)
go-selinux/selinux_linux.go:            if err != unix.EINTR { //nolint:errorlint // unix errors are bare
go-selinux/selinux_linux.go:    return filepath.Join("/proc/self/task", strconv.Itoa(unix.Gettid()), "attr", attr)
go-selinux/selinux_linux.go:    l, err := unix.GetsockoptString(int(fd), unix.SOL_SOCKET, unix.SO_PEERSEC)
go-selinux/xattrs_linux.go:     for errno == unix.ERANGE { //nolint:errorlint // unix errors are bare
go-selinux/xattrs_linux.go:             sz, err := unix.Lgetxattr(path, attr, dest)
go-selinux/xattrs_linux.go:             if err != unix.EINTR { //nolint:errorlint // unix errors are bare
go-selinux/xattrs_linux.go:     for errno == unix.ERANGE { //nolint:errorlint // unix errors are bare
go-selinux/xattrs_linux.go:             sz, err := unix.Getxattr(path, attr, dest)
go-selinux/xattrs_linux.go:             if err != unix.EINTR { //nolint:errorlint // unix errors are bare
```

kolyshkin:
OK, since we only use x/sys/linux, this doesn't make sense.

PR #194, opened 2022-11-02
go-selinux: add coverage for LfileLabel, LsetFileLabel

thaJeztah:
updates: 1b18907dde2de75c9762d425c862586e1f7c438a (https://github.com/opencontainers/selinux/pull/169)

### DIFF
--- a/go-selinux/label/label_stub_test.go
+++ b/go-selinux/label/label_stub_test.go
@@ -79,11 +79,12 @@ func TestCheckLabelCompile(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if _, err := FileLabel("/etc"); err != nil {
+	tmpDir := t.TempDir()
+	if _, err := FileLabel(tmpDir); err != nil {
 		t.Fatal(err)
 	}
 
-	if err := SetFileLabel("/etc", "foobar"); err != nil {
+	if err := SetFileLabel(tmpDir, "foobar"); err != nil {
 		t.Fatal(err)
 	}
 
--- a/go-selinux/selinux_linux_test.go
+++ b/go-selinux/selinux_linux_test.go
@@ -16,24 +16,65 @@ func TestSetFileLabel(t *testing.T) {
 		t.Skip("SELinux not enabled, skipping.")
 	}
 
-	tmp := "selinux_test"
-	con := "system_u:object_r:bin_t:s0"
-	out, err := os.OpenFile(tmp, os.O_WRONLY|os.O_CREATE, 0)
+	const (
+		tmpFile = "selinux_test"
+		tmpLink = "selinux_test_link"
+		con     = "system_u:object_r:bin_t:s0:c1,c2"
+		con2    = "system_u:object_r:bin_t:s0:c3,c4"
+	)
+
+	_ = os.Remove(tmpFile)
+	out, err := os.OpenFile(tmpFile, os.O_WRONLY|os.O_CREATE, 0)
 	if err != nil {
-		t.Fatalf("unable to open %s: %s", tmp, err)
+		t.Fatal(err)
 	}
 	out.Close()
-	defer os.Remove(tmp)
+	defer os.Remove(tmpFile)
 
-	if err := SetFileLabel(tmp, con); err != nil {
+	_ = os.Remove(tmpLink)
+	if err := os.Symlink(tmpFile, tmpLink); err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(tmpLink)
+
+	if err := SetFileLabel(tmpLink, con); err != nil {
 		t.Fatalf("SetFileLabel failed: %s", err)
 	}
-	filelabel, err := FileLabel(tmp)
+	filelabel, err := FileLabel(tmpLink)
 	if err != nil {
 		t.Fatalf("FileLabel failed: %s", err)
 	}
-	if con != filelabel {
+	if filelabel != con {
 		t.Fatalf("FileLabel failed, returned %s expected %s", filelabel, con)
+	}
+
+	// Using LfileLabel to verify that the symlink itself is not labeled.
+	linkLabel, err := LfileLabel(tmpLink)
+	if err != nil {
+		t.Fatalf("LfileLabel failed: %s", err)
+	}
+	if linkLabel == con {
+		t.Fatalf("Label on symlink should not be set, got: %q", linkLabel)
+	}
+
+	// Use LsetFileLabel to set a label on the symlink itself.
+	if err := LsetFileLabel(tmpLink, con2); err != nil {
+		t.Fatalf("LsetFileLabel failed: %s", err)
+	}
+	filelabel, err = FileLabel(tmpFile)
+	if err != nil {
+		t.Fatalf("FileLabel failed: %s", err)
+	}
+	if filelabel != con {
+		t.Fatalf("FileLabel was updated, returned %s expected %s", filelabel, con)
+	}
+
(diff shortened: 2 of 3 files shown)


rhatdan:
LGTM

go-selinux/selinux_linux_test.go
```diff
@@ -16,25 +16,61 @@ func TestSetFileLabel(t *testing.T) {
...
+	if filelabel != con {
 		t.Fatalf("FileLabel failed, returned %s expected %s", filelabel, con)
 	}
+
+	// Using LfileLabel to verify that the symlink itself is not labeled.
+	linkLabel, err := LfileLabel(tmpLink)
+	if err != nil {
+		t.Fatalf("LfileLabel failed: %s", err)
+	}
+	if linkLabel != "" {
```
rhatdan, line 51: This should be linkLabel == con

go-selinux/selinux_stub_test.go
```diff
@@ -22,6 +22,13 @@ func TestSELinuxStubs(t *testing.T) {
 		t.Error(err)
 	}
 
+	if _, err := LfileLabel("/etc"); err != nil {
+		t.Error(err)
+	}
+	if err := LsetFileLabel("/etc", testLabel); err != nil {
```
rhatdan, line 28: This could be dangerous if run as root.

thaJeztah: Yes, I was looking for a good path to use (this is for the stubs, so _should_ not do anything, but the existing tests also used `/etc`)

go-selinux/selinux_linux_test.go
thaJeztah: My thinking here was that no label should be set _at all_ on the symlink, correct?

thaJeztah:
(oh! forgot to post); Hmm... right, so this won't help a lot; while looking, of course I stumbled upon this ticket from @kolyshkin 😂  https://github.com/actions/runner-images/issues/2307

Also had a look if we could install SELinux on Ubuntu, but it requires a reboot, so not an option 😞 

```
=== RUN   TestSetFileLabel
    selinux_linux_test.go:16: SELinux not enabled, skipping.
--- SKIP: TestSetFileLabel (0.00s)
=== RUN   TestKVMLabels
    selinux_linux_test.go:78: SELinux not enabled, skipping.
--- SKIP: TestKVMLabels (0.00s)
=== RUN   TestInitLabels
    selinux_linux_test.go:99: SELinux not enabled, skipping.
--- SKIP: TestInitLabels (0.00s)
=== RUN   TestSELinux
    selinux_linux_test.go:131: SELinux not enabled, skipping.
--- SKIP: TestSELinux (0.00s)
...
```

go-selinux/selinux_stub_test.go
thaJeztah: Updated to use a temp-file to test with 👍

go-selinux/selinux_linux_test.go
rhatdan: All files have labels.  Getting a file without a label on an SELinux system is difficult.

go-selinux/selinux_stub_test.go
rhatdan: If we are using system directories elsewhere we probably should not.

rhatdan:
We really should be testing on Centos or Fedora.

rhatdan:
Apply this patch to yours and the tests passes.
[diff.patch](https://github.com/opencontainers/selinux/files/9923538/diff.patch.txt)

kolyshkin:
> Also had a look if we could install SELinux on Ubuntu, but it requires a reboot, so not an option

I remember @avagin implemented a complicated trick with upgrading the kernel and rebooting a travis machine while keeping the process that connects to travis infra intact (using CRIU), which helped us to test newer kernels.

Today, we can use something like cirrus-ci which you can use to install centos (7, 8, or 9) and also run fedora via KVM. See runc's .cirrus.yml for example.

We can also use github ci to run fedora in a KVM, but this is more error prone since the only way is to use Mac OS X and there's a 5 to 10% chance that it will fail, based on what I saw.

go-selinux/label/label_stub_test.go
```diff
@@ -75,15 +79,22 @@ func TestProcessLabel(t *testing.T) {
 }
 
 func TestCheckLabelCompile(t *testing.T) {
+	tmpFile := filepath.Join(t.TempDir(), "test_file")
+	out, err := os.OpenFile(tmpFile, os.O_WRONLY|os.O_CREATE, 0)
+	if err != nil {
+		t.Fatal(err)
+	}
```
kolyshkin, line 86: Why not merely use `t.TempDir()`. It returns a directory name which you can use directly as an argument to `FileLabel` and `SetFileLabel`.

go-selinux/selinux_stub_test.go
```diff
@@ -13,12 +15,18 @@ func TestSELinuxStubs(t *testing.T) {
 	if GetEnabled() {
 		t.Error("SELinux enabled on non-linux.")
 	}
+	tmpFile := filepath.Join(t.TempDir(), "test_file")
```
kolyshkin, line 18: ditto

kolyshkin (changes requested):
A nits to simplify code by using t.TempDir more directly

rhatdan:
@thaJeztah Still working on this?

thaJeztah:
Oh! Looks like I somehow missed the comments here; updated 👍

go-selinux/label/label_stub_test.go
```diff
@@ -3,7 +3,9 @@
 
 package label
 
-import "testing"
+import (
+	"testing"
+)
```
thaJeztah, line 8: Oh! Let me revert this part

thaJeztah: done 👍

rhatdan:
LGTM
@kolyshkin @giuseppe @vrothberg PTAL

rhatdan:
@lsm5 PTAL

rhatdan (approved):

thaJeztah:
> @lsm5 PTAL

Oh! Voice from the past. It's been a while since we interacted (hope you're doing well!)

rhatdan:
@lsm5 has shown interest in SELinux so I figured I would ping him.

lsm5:
> > @lsm5 PTAL
> 
> Oh! Voice from the past. It's been a while since we interacted (hope you're doing well!)

@thaJeztah :wave: I'm good, hope you are too! :smile:

> @lsm5 has shown interest in SELinux so I figured I would ping him.

@rhatdan shown interest yes, but this is outta my reviewing capacity atm, LGTM from a cursory look though. Also `sudo make test` passes for this PR on my F37 env.

Also, RE: fedora/centos testing, do you want that to be a CI task here?

rhatdan:
It would be awesome if we could test Fedora or Centos here.

lsm5:
@rhatdan IIUC, github actions is ubuntu only. So, we'll likely need some other setup. 

@cevich wdyt? will it be a lot of pain to enable cirrus outside the containers org?

thaJeztah:
Yeah, it's complicated. It's possible to attach custom runners to GitHub actions, but those would need to be maintained somewhere (_possibly_ the CNCF has infra for this?). Some caution is needed for these as well to prevent those runners from also be running on forks (and to be sure they're only running on "good" pull requests, to prevent crypto miners from using them).

Of course the ideal would be for GitHub to provide at least "one" variant of an rpm/yum/dnf flavoured distro.

PR #200, opened 2023-07-27
fix some error

ningmingxiao:


go-selinux/selinux_linux.go
```diff
@@ -1134,9 +1134,7 @@ func rchcon(fpath, label string) error { //revive:disable:cognitive-complexity
 	}
 	return pwalkdir.Walk(fpath, func(p string, _ fs.DirEntry, _ error) error {
 		if fastMode {
-			if cLabel, err := lFileLabel(fpath); err == nil && cLabel == label {
```
thaJeztah, line 1137: Was this perhaps meant to be

```suggestion
			if cLabel, err := lFileLabel(p); err == nil && cLabel == label {
```

Because I think removing this block entirely would mean `pwalkdir.Walk()` would not be walking anything? (so the whole `rchcon()` could return directly at line 1133? 🤔

ningmingxiao:
ping @thaJeztah

ningmingxiao:
@thaJeztah can this pr be ok?

rhatdan:
@giuseppe PTAL

giuseppe (approved):
LGTM

PR #205, opened 2023-08-26
ci: bump some actions

kolyshkin:
1. Bump actions/setup-go to v4 which enables caching by default. Disable cache for cases when golangci-lint is used.

2. Remove 'stable:' property from actions/setup-go (not needed since at least v3).

3. Bump tim-actions/get-pr-commits to v1.3.0  (fixes node compatibility).


rhatdan (approved):
LGTM

thaJeztah (approved):
LGTM, thanks!

PR #206, opened 2023-08-29
Misc nitpicks

kolyshkin:
1. Fix Makefile.
2. Add TODO to remove min/max.
3. Add codespell to CI.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -39,6 +39,16 @@ jobs:
         with:
           version: v1.54
 
+  codespell:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v3
+    - name: install deps
+      # Version of codespell bundled with Ubuntu is way old, so use pip.
+      run: pip install codespell
+    - name: run codespell
+      run: codespell
+
   cross:
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -25,7 +25,7 @@ build-cross:
 
 .PHONY: test
 test:
-	go test -timeout 3m ${TESTFLAGS} -v ./...
+	$(GO) test -timeout 3m ${TESTFLAGS} -v ./...
```
thaJeztah, line 28: I'm not sure if overriding versions still work with their recent changes to automatically install versions and downgrade language versions based on go.mod 🤔

See https://tip.golang.org/doc/toolchain

.github/workflows/validate.yml
```diff
@@ -39,6 +39,16 @@ jobs:
...
+  codespell:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v3
+    - name: install deps
+      # Version of codespell bundled with Ubuntu is way old, so use pip.
+      run: pip install codespell
+    - name: run codespell
+      run: codespell
+
```
thaJeztah, line 51: Oh! I wanted to check if there was an action for this; and  then I stumbled on a PR that was adding it; so: yes, there's an action for this; https://github.com/codespell-project/actions-codespell so we may as well use that?

rhatdan:
LGTM

PR #210, opened 2024-02-13
Add dependabot config

kolyshkin:
This way it can create PRs to update various dependencies (in Go modules and github workflow rules).

Cc @rhatdan

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  # Dependencies listed in .github/workflows/*.yml
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


austinvazquez (approved):

austinvazquez:
LGTM, another thread I would like to start is enabling rubberstamp approvals and auto-merge automation for minor dependency updates. Dependabot can create quite a few PRs and overwhelm maintainers. I can try a proof of concept and propose a mechanism separately from this.

.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  # Dependencies listed in go.mod
+  - package-ecosystem: "gomod"
```
thaJeztah, line 7: With this repository being a library, I don't think we need to update module versions, unless there's a critical reason for it in the call path. Doing so would violate go modules MVS (which requires lowest version to be specified)

thaJeztah (changes requested):

.github/dependabot.yml
kolyshkin: Right. I'll remove this.

(Wonder what to do about other repos such as runc, which is a library and a binary at the same time)

kolyshkin:
> LGTM, another thread I would like to start is enabling rubberstamp approvals and auto-merge automation for minor dependency updates. Dependabot can create quite a few PRs and overwhelm maintainers. I can try a proof of concept and propose a mechanism separately from this. Perhaps we can start with selinux and if successful apply the mechanism to other projects? cc: @kolyshkin @rhatdan wdyt?

Not a big deal for this repo, as we don't have lots of dependencies and we don't require 2 LGTMs here, so any single maintainer can merge.

OTOH I'd love some automation for github.com/opencontainers/runc (which requires two LGTMs, while we can easily do with just one for dependabot and other simple PRs, but I don't know how to do that).

.github/dependabot.yml
thaJeztah: Yeah, those can be the tricky ones. I think currently it's probably OK (we update main, but not release branches), although we could be more conservative if the updates we pull in bring no changes to code used in runc itself

thaJeztah (approved):
LGTM

updating actions sounds good to me, and shouldn't hurt. Let's see if the "daily" works well

rhatdan:
LGTM

PR #214, opened 2024-07-29
skip write /proc/self/attr/keycreate when selinux label is null

ningmingxiao:
https://github.com/opencontainers/runc/pull/4354/files

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -724,11 +724,11 @@ func peerLabel(fd uintptr) (string, error) {
 // setKeyLabel takes a process label and tells the kernel to assign the
 // label to the next kernel keyring that gets created
 func setKeyLabel(label string) error {
-	err := writeCon("/proc/self/attr/keycreate", label)
-	if errors.Is(err, os.ErrNotExist) {
+	if label == "" {
 		return nil
 	}
-	if label == "" && errors.Is(err, os.ErrPermission) {
+	err := writeCon("/proc/self/attr/keycreate", label)
+	if errors.Is(err, os.ErrNotExist) {
 		return nil
 	}
 	return err


ningmingxiao:
@kolyshkin

thaJeztah:
Looks like your commit is missing a DCO sign-off in the commit message; can you add one? When doing so, please use your real name for the sign-off (not your github handle/username); you can find some instructions in the output from the DCO check here; https://github.com/opencontainers/selinux/pull/214/checks?check_run_id=28026773649

ningmingxiao:
> Looks like your commit is missing a DCO sign-off in the commit message; can you add one? When doing so, please use your real name for the sign-off (not your github handle/username); you can find some instructions in the output from the DCO check here; https://github.com/opencontainers/selinux/pull/214/checks?check_run_id=28026773649

done

thaJeztah:
Thanks! Oh minor nit; `label` is a `string`, but the commit message says `when selinux label is null`

Perhaps;

```
SetKeyLabel: ignore empty selinux labels
```


Looking at the change; it looks like we don't have a test-case for setting an empty label; would probably be good to add one; https://github.com/opencontainers/selinux/blob/bb1ec25c889f29e35dffec33346af3c7d434dbc1/go-selinux/label/label_linux_test.go#L197-L211

thaJeztah:
In general; I see that previously we would try to set an empty label, but then ignore the error if it was a `os.ErrPermission`.

Slightly wondering, but maybe out of scope for this PR (and I could use input from @kolyshkin and @rhatdan on this);

- Should (ideally?) empty labels be handled by the caller (i.e. don't call `SetKeyLabel` if there's no label)
- And instead `SetKeyLabel` returning an error (invalid label)?

Somewhat related to the above; could there be valid use-cases of trying to unset this? (so could an empty label have the intent of "unset" what's there?). I see you linked to a runc PR;

- https://github.com/opencontainers/runc/pull/4354

Looking at that PR;

- I think it might be more _correct_ for the caller (runc) to not call `SetKeyLabel` if there's no value to set?
- Also looking at the code, I see various instances of `defer selinux.SetKeyLabel("")` - do we know what the intent there was? Was the intent to reset values that were previously set?

Because if that's the case, then possibly this PR is actually hiding a bug / incorrect use in runc, and it would be better for `SetKeyLabel` to return an error on empty values.

Screenshot 2024-07-29 at 11 20 42

ningmingxiao:
we should let runc do it

PR #217, opened 2024-10-16
VERSION: remove

kolyshkin:
This is not used anywhere, and go tools use git tags anyway.


kolyshkin:
The reason for this I just foolishly published a new selinux release without bumping a VERSION file.

thaJeztah (approved):
LGTM

there's no binary releases of this module, so yeah, doesn't provide much IMO

rhatdan (approved):
LGTM

PR #218, opened 2024-10-16
Bump Go version, deps, fix some linter issues...

kolyshkin:
This started as a bump of some CI deps but quickly got out of hand 😬 

See individual commits for details. High level overview:
- ci: bump ubuntu to 24.04 (see https://github.com/actions/runner-images/issues/11101);
- ~~go.mod: bump go to 1.21, drop min/max functions;~~
- ci: add go 1.19, 1.23, 1.24, drop go 1.21, 1.22 from test matrix;
- fix or silence new gosec warnings (these commits need to be reviewed carefully);
- ci: bump golangci-lint to v1.64, fix golangci-lint jobs caching;
- ci: add 32-bit tests;
- ci: add `all-done` job;
- go.mod: bump to `golang.org/x/sys v0.1.0`;
- ci: show avc denials after tests.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -126,3 +126,6 @@ jobs:
 
       - name: "make test"
         run: lima make -C /tmp/selinux test
+
+      - name: "32-bit test"
+        run: lima make -C /tmp/selinux GOARCH=386 test
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -10,7 +10,7 @@ on:
 jobs:
 
   commit:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     # Only check commits on pull requests.
     if: github.event_name == 'pull_request'
     steps:
@@ -28,7 +28,7 @@ jobs:
           error: 'Subject too long (max 72)'
 
   lint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
@@ -40,7 +40,7 @@ jobs:
           version: v1.56
 
   codespell:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v4
     - name: install deps
@@ -50,7 +50,7 @@ jobs:
       run: codespell
 
   cross:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
       - name: cross
@@ -76,7 +76,7 @@ jobs:
       matrix:
         go-version: [1.21.x, 1.22.x]
         race: ["-race", ""]
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -34,10 +34,9 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version: 1.22.x
-          cache: false # golangci-lint-action does its own caching
       - uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.56
+          version: v1.64
 
   codespell:
     runs-on: ubuntu-24.04
@@ -63,10 +62,9 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version: 1.22.x
-          cache: false # golangci-lint-action does its own caching
       - uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.56
+          version: v1.64
       - name: test-stubs
         run: make test
 
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -72,7 +72,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.21.x, 1.22.x]
+        go-version: [1.19.x, 1.23.x, 1.24.x]
         race: ["-race", ""]
     runs-on: ubuntu-24.04
     steps:
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -125,10 +125,15 @@ jobs:
           lima sudo dnf install --setopt=install_weak_deps=false --setopt=tsflags=nodocs -y git-core make golang
 
       - name: "make test"
+        continue-on-error: true
         run: lima make -C /tmp/selinux test
 
       - name: "32-bit test"
+        continue-on-error: true
         run: lima make -C /tmp/selinux GOARCH=386 test
+
+      - name: "Show AVC denials"
+        run: lima sudo ausearch -m AVC,USER_AVC || true
 
   all-done:
     needs:
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -129,3 +129,16 @@ jobs:
 
       - name: "32-bit test"
         run: lima make -C /tmp/selinux GOARCH=386 test
+
+  all-done:
+    needs:
+      - commit
+      - lint
+      - codespell
+      - cross
+      - test-stubs
+      - test
+      - vm
+    runs-on: ubuntu-24.04
+    steps:
+    - run: echo "All jobs completed"
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -584,7 +584,8 @@ func bitsetToStr(c *big.Int) string {
 	var str string
 
 	length := 0
-	for i := int(c.TrailingZeroBits()); i < c.BitLen(); i++ {
+	i0 := int(c.TrailingZeroBits()) //#nosec G115 -- don't expect TralingZeroBits to return values with highest bit set.
+	for i := i0; i < c.BitLen(); i++ {
 		if c.Bit(i) == 0 {
 			continue
 		}
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -138,6 +138,7 @@ func verifySELinuxfsMount(mnt string) bool {
 		return false
 	}
 
+	//#nosec G115 -- there is no overflow here.
 	if uint32(buf.Type) != uint32(unix.SELINUX_MAGIC) {
 		return false
 	}
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -45,7 +45,7 @@ type selinuxState struct {
 
 type level struct {
 	cats *big.Int
-	sens uint
+	sens int
 }
 
 type mlsRange struct {
@@ -501,14 +501,14 @@ func catsToBitset(cats string) (*big.Int, error) {
 				return nil, err
 			}
 			for i := catstart; i <= catend; i++ {
-				bitset.SetBit(bitset, int(i), 1)
+				bitset.SetBit(bitset, i, 1)
 			}
 		} else {
 			cat, err := parseLevelItem(ranges[0], category)
 			if err != nil {
 				return nil, err
 			}
-			bitset.SetBit(bitset, int(cat), 1)
+			bitset.SetBit(bitset, cat, 1)
 		}
 	}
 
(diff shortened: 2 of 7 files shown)


kolyshkin:
Oops, apparently most tests are skipped (because, of course, Ubuntu does not have selinux).

```
=== RUN   TestSetFileLabel
    selinux_linux_test.go:16: SELinux not enabled, skipping.
--- SKIP: TestSetFileLabel (0.00s)
=== RUN   TestKVMLabels
    selinux_linux_test.go:83: SELinux not enabled, skipping.
--- SKIP: TestKVMLabels (0.00s)
=== RUN   TestInitLabels
    selinux_linux_test.go:104: SELinux not enabled, skipping.
--- SKIP: TestInitLabels (0.00s)
=== RUN   TestSELinux
    selinux_linux_test.go:136: SELinux not enabled, skipping.
--- SKIP: TestSELinux (0.00s)
=== RUN   TestSetEnforceMode
    selinux_linux_test.go:183: SELinux not enabled, skipping.
--- SKIP: TestSetEnforceMode (0.00s)
=== RUN   TestCanonicalizeContext
    selinux_linux_test.go:206: SELinux not enabled, skipping.
--- SKIP: TestCanonicalizeContext (0.00s)
=== RUN   TestFindSELinuxfsInMountinfo
    selinux_linux_test.go:270: found "/sys/fs/selinux"
    selinux_linux_test.go:270: found "/root/chroot/selinux"
    selinux_linux_test.go:270: found ""
--- PASS: TestFindSELinuxfsInMountinfo (0.00s)
=== RUN   TestSecurityCheckContext
    selinux_linux_test.go:279: SELinux not enabled, skipping.
--- SKIP: TestSecurityCheckContext (0.00s)
=== RUN   TestClassIndex
    selinux_linux_test.go:304: SELinux not enabled, skipping.
--- SKIP: TestClassIndex (0.00s)
=== RUN   TestComputeCreateContext
    selinux_linux_test.go:324: SELinux not enabled, skipping.
....
```

Need to switch to a hosted Fedora or something like that. But that's a separate issue.

Let this be draft for now.

kolyshkin:
> Oops, apparently most tests are skipped (because, of course, Ubuntu does not have selinux).

Of course [we knew it for a long time](https://github.com/opencontainers/selinux/pull/194#issuecomment-1300420091), we just forgot :)

And since https://github.com/actions/runner-images/issues/2307 is quite old, I filed a new one: https://github.com/actions/runner-images/issues/10802

rhatdan:
LGTM, Are the SELinux tests running on Ubuntu now?

kolyshkin:
> LGTM, Are the SELinux tests running on Ubuntu now?

No :(

We need to switch to Fedora, I filed https://github.com/opencontainers/selinux/issues/220 for that.

This PR still makes sense but with or without it we actually don't test much :(

alexellis:
Current generations of arm64 servers available via cloud do not have working nested virtualisation. It may come in the future, where a standard runner could pivot into a RPM-based distro.

So whilst that's not available, and GitHub may not have it on their radar to support SELinux, we could provide SELinux based runners with fedora or a fedora-like distro like Alma for x86_64 and arm64.

Feel free to reach out to me to discuss options if you know of an end-user company willing to sponsor it.

kolyshkin:
@rhatdan @thaJeztah PTAL (the lack of selinux testing is now addressed by #221).

rhatdan:
It would be helpful if the audit.log is collected on failures, to see if there is an AVC reported.

kolyshkin:
> It would be helpful if the audit.log is collected on failures, to see if there is an AVC reported.

If we're talking about TestKeyLabel failures, I can easily reproduce it locally on Fedora 41 and I never see any new denials entries in audit.log. But let's discuss it here: https://github.com/opencontainers/selinux/issues/222.

If we're talking about this PR, is it sufficient to collect the output of `sudo ausearch -m AVC` after the tests? If yes, I just added this step as a last commit in this PR. If no, please suggest an alternative.

.github/workflows/validate.yml
```diff
@@ -125,11 +125,16 @@ jobs:
...
       - name: "make test"
+        continue-on-error: true
         run: lima make -C /tmp/selinux test
 
       - name: "32-bit test"
+        continue-on-error: true
         run: lima make -C /tmp/selinux GOARCH=386 test
 
+      - name: "Show AVC denials"
+        run: lima sudo ausearch -m AVC || true
```
rhatdan, line 136: -m AVC,USER_AVC

kolyshkin:
> If you run the test locally in permissive mode, does it still blow up? If yes, then the issue is DAC, if no, then the issue is SELinux, and the AVC must be blocked via a dontaudit rule.

Either `permissive` or `enforcing`, it blows up the same way:
<details>

```console
[kir@kir-tp1 selinux]$ sestatus
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   enforcing
Mode from config file:          permissive
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Memory protection checking:     actual (secure)
Max kernel policy version:      33

[kir@kir-tp1 selinux]$ go test -count 100 -run KeyLabel ./go-selinux/label/ | grep -c FAIL
0

[kir@kir-tp1 selinux]$ go test -count 10000 -run KeyLabel ./go-selinux/label/ | grep -c FAIL
6719

[kir@kir-tp1 selinux]$ sudo setenforce 0
[sudo] password for kir: 

[kir@kir-tp1 selinux]$ go test -count 10000 -run KeyLabel ./go-selinux/label/ | grep -c FAIL
5912

[kir@kir-tp1 selinux]$ go test -count 10000 -run KeyLabel ./go-selinux/label/ | grep -c FAIL
6788

[kir@kir-tp1 selinux]$ sestatus 
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   permissive
Mode from config file:          permissive
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Memory protection checking:     actual (secure)
Max kernel policy version:      33

```
</details>

.github/workflows/validate.yml
kolyshkin: done

rhatdan:
So this is a DAC issue, and the key is not writable from a non-root user, would be my guess.

kolyshkin:
> So this is a DAC issue, and the key is not writable from a non-root user, would be my guess.

@rhatdan this way it would *always* fail, but what we see is it fails randomly.

Also, it fails in the same way (i.e. randomly) when the test run as a root user (with `CAP_DAC_OVERRIDE` et al):
```console
[kir@kir-tp1 selinux]$ pwd
/home/kir/git/opencontainers/selinux
[kir@kir-tp1 selinux]$ cd go-selinux/label/
[kir@kir-tp1 label]$ go test -c .
[kir@kir-tp1 label]$ sudo ./label.test -test.run KeyLabel -test.count 10000 | grep -c FAIL
6205
```

(note I'm running 10000 iterations of the test, of which it fails 6205 times).

A (presumably) similar C program do not fail.

rhatdan:
Ok @kolyshkin has diagnosed this to be a GOLang thread issue.

alexellis:
I'm going to unsubscribe to this thread, but if anyone wants to take us up on the offer to add an SELinux/RPM-based distro as a runner, please reach out via email.

kolyshkin:
Flakes are being fixed by #227; I'd rather merge this PR first though (before #227).

go-selinux/selinux_linux.go
```diff
@@ -639,21 +639,6 @@ func (m mlsRange) String() string {
 	return low + "-" + high
 }
 
-// TODO: remove min and max once Go < 1.21 is not supported.
-func max(a, b uint) uint {
```
thaJeztah, line 643: Was there a strong reason not to continue supporting older Go versions? If not, I'd prefer continuing compatibility; we can either move this to a `//go:build ...` file, or rename the func to not shadow the builtin

go.mod
```diff
@@ -1,5 +1,5 @@
 module github.com/opencontainers/selinux
 
-go 1.19
+go 1.21
```
thaJeztah, line 3: Same here, unless we really cannot

go-selinux/selinux_linux.go
```diff
@@ -138,7 +138,7 @@ func verifySELinuxfsMount(mnt string) bool {
 		return false
 	}
 
-	if uint32(buf.Type) != uint32(unix.SELINUX_MAGIC) {
+	if uint32(buf.Type) != uint32(unix.SELINUX_MAGIC) { //nolint:gosec
```
thaJeztah, line 141: Can you use the `#nosec G115` style? That way we can supported only `G115` without any other gosec issue that may be found

go-selinux/selinux_linux.go
```diff
@@ -583,7 +583,8 @@ func bitsetToStr(c *big.Int) string {
 	var str string
 
 	length := 0
-	for i := int(c.TrailingZeroBits()); i < c.BitLen(); i++ {
+	i0 := int(c.TrailingZeroBits()) //nolint:gosec
```
thaJeztah, line 586: Same here, and possible add a comment after the `// #nosec G115 -- <short commend here >`

go-selinux/selinux_linux.go
kolyshkin: No strong reasons; the weak ones are quite usual:
 - by switching to a newer version we can use new language and stdlib features, some of those I really like;
 - Go < 1.22 is not supported and might have some unpatched issues.

Sure, if there is a need, we can try maintaining backward compatibility, especially since this repo is not too bif or complex.

go-selinux/selinux_linux.go
kolyshkin: Alas this is not working for me. Perhaps you only used standalone `gosec`?

```console
$ golangci-lint run ./...
go-selinux/selinux_linux.go:141:11: G115: integer overflow conversion int64 -> uint32 (gosec)
	if uint32(buf.Type) != uint32(unix.SELINUX_MAGIC) { // #nosec G115 -- this code is right.
	         ^
```

go-selinux/selinux_linux.go
kolyshkin: > we can either move this to a //go:build ... file

It's not possible. Now I remember I actually tried that before, only to get something like this:

```console
$ make
GOOS=linux GOARCH=amd64 go build  ./...
# github.com/opencontainers/selinux/go-selinux
go-selinux/selinux_linux.go:666:22: built-in max requires go1.21 or later (-lang was set to go1.19; check go.mod)
go-selinux/selinux_linux.go:669:23: built-in min requires go1.21 or later (-lang was set to go1.19; check go.mod)
```

go-selinux/selinux_linux.go
kolyshkin: This is weird. The same annotation works just fine in a different place:

```go
i0 := int(c.TrailingZeroBits()) //#nosec G115 -- don't expect TralingZeroBits to return values with highest bit set.
```

bit not here. I am lost 😿

kolyshkin: OK, it doesn't work because it's after `{`. It works if I put in on a previous line:

```go
       //#nosec G115 -- there is no overflow here.
       if uint32(buf.Type) != uint32(unix.SELINUX_MAGIC) {
```

kolyshkin: But, in this case it might want for the whole block, rather than just this line. I'm lost again 😕

kolyshkin: Ah, the "whole block" is `return false` so I guess it's fine.

kolyshkin:
@thaJeztah please take another look, I think I've addressed all your comments.

Aside from changes that you've requested, I added Go 1.19 to CI matrix, and added a new commit which removes a couple of unneeded `nolint:gosec` annotations.

go-selinux/selinux_linux.go
kolyshkin: Done

go-selinux/selinux_linux.go
kolyshkin: Would like to hear your reasons for supporting Go 1.19.

go-selinux/selinux_linux.go
thaJeztah: Yeah, they are kinda horrible, the `#nosec` comments! I think it's good to use them though (to not accidentally mask other security-related discoveries).

Thanks!

thaJeztah:
Thanks! Yeah, users of older go versions should become rare, but as long as it's not adding a ton of burden (generally, this repository won't require too many updates I expect), I think  it's good to keep `<oldest supported>`, `current stable`, `old stable` as our test-matrix.

But we sure can update `<oldest supported>` once things become complicated / we _significantly_ benefit from new go features.

thaJeztah (approved):
LGTM, thanks!

kolyshkin:
> Thanks! Yeah, users of older go versions should become rare, but as long as it's not adding a ton of burden (generally, this repository won't require too many updates I expect), I think it's good to keep `<oldest supported>`, `current stable`, `old stable` as our test-matrix.

Yes to that!

> 
> But we sure can update `<oldest supported>` once things become complicated / we _significantly_ benefit from new go features.

In this particular case this was min/max functions, which I had to adopt in two regards:
 - change from uint to int (as their user changed the type);
 - change their names (or add linter annotations).
 
 I played a bit and decided the best way forward is to just bump min Go version and drop them.
 
 But it's not a big deal to keep compatibility if needed. At least for now.

rhatdan:
Might be worth adding renovate to this repo to keep the repo more up 2 date.

thaJeztah:
> Might be worth adding renovate to this repo to keep the repo more up 2 date.

For github actions used, and go version, probably. For go.mod, I think we can stick to go's MVS recommendations (specify the minimum version required)

PR #219, opened 2024-10-17
[test] test in actuated runners

kolyshkin:
(currently includes #218 but it's not really required here).

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -37,7 +37,7 @@ jobs:
           cache: false # golangci-lint-action does its own caching
       - uses: golangci/golangci-lint-action@v6
         with:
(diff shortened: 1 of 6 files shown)


kolyshkin:
> @rhatdan can you take a look? (I've seen occasional errors from `TestKeyLabel` locally, others look new to me).

Nevermind, [most of] the failures are caused by missing container-selinux (and thus things like `container_t` and `container_file_t`), which is not available on Ubuntu. Can try to install it manually, but overall I guess it is easier to test on an RHEL clone instead.

kolyshkin:
Going with Cirrus CI instead, but I'm stuck as I don't have rights.
https://github.com/opencontainers/selinux/issues/220

PR #221, opened 2025-02-25
CI: add AlmaLinux 8, CentOS Stream 9, and Fedora

### LINKED ISSUE #220: Set up Cirrus CI to test on Fedora
Since GHA only supports Ubuntu Linux, which does not come with SELinux enabled, CI setup for this repository is currently useless (see [here](https://github.com/opencontainers/selinux/pull/194#issuecomment-1300420091) or [here](https://github.com/opencontainers/selinux/pull/218#issuecomment-2418197893)).

I propose to add `.cirrus.yml` and run some tests on RHEL clones (and, maybe, Fedora), as they come with SELinux already configured.

We need some help configuring Cirrus CI integration for this repo.

~~(will add more details later, consider this a draft for now)~~

AkihiroSuda:
Lima (http://lima-vm.io) is used to create VM of these distros.

Close #220

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -90,3 +90,41 @@ jobs:
 
       - name: test
         run: make TESTFLAGS="${{ matrix.race }}" test
+
+  vm:
+    name: "VM"
+    strategy:
+      fail-fast: false
+      matrix:
+        template:
+          - template://almalinux-8
+          - template://centos-stream-9
+          - template://fedora
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: "Install Lima"
+        uses: lima-vm/lima-actions/setup@v1
+        id: lima-actions-setup
+
+      - name: "Cache ~/.cache/lima"
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/lima
+          key: lima-${{ steps.lima-actions-setup.outputs.version }}-${{ matrix.template }}
+
+      - name: "Start VM"
+        # --plain is set to disable file sharing, port forwarding, built-in containerd, etc. for faster start up
+        run: limactl start --plain --name=default ${{ matrix.template }}
+
+      - name: "Initialize VM"
+        run: |
+          set -eux -o pipefail
+          # Sync the current directory to /tmp/selinux in the guest
+          limactl cp -r . default:/tmp/selinux
+          # Install packages
+          lima sudo dnf install -y git make golang
+
+      - name: "make test"
+        run: lima make -C /tmp/selinux test


kolyshkin (dismissed):
Thank you! LGTM

mrunalp (approved):

AkihiroSuda:
Updated PR to use https://github.com/lima-vm/lima-actions

kolyshkin (approved):
Still LGTM

.github/workflows/validate.yml
```diff
@@ -90,3 +90,41 @@ jobs:
...
+
+      - name: "Install Lima"
+        uses: lima-vm/lima-actions/setup@v1
+        id: lima-actions-setup
+
+      - name: "Cache ~/.cache/lima"
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/lima
+          key: lima-${{ steps.lima-actions-setup.outputs.version }}
```
kolyshkin, line 115: Do I understand it right that this step caches and restores the image downloaded by `limactl start`?

If yes, the image name should probably be the key here, not the lima version. Or, if you foresee that the download format will change drastically soon, both the lima version (perhaps just a major version?) and the image name.

kolyshkin: One other observation, such caching can be internal to `lima-actions/setup`.

But that would not work for the case when there are different images. Hmm.

kolyshkin (changes requested):
A few questions re: caching.

.github/workflows/validate.yml
AkihiroSuda: Added the image name to the key

> Or, if you foresee that the download format will change drastically soon, both the lima version (perhaps just a major version?) and the image name.

It is not a matter of the format, it is a matter of the qcow2 digest in [`/usr/local/share/lima/templates/fedora.yaml`](https://github.com/lima-vm/lima/blob/v1.0.6/templates/fedora.yaml)

AkihiroSuda:
Flaky test on Fedora
```
=== RUN   TestKeyLabel
    label_linux_test.go:202: write /proc/self/attr/keycreate: permission denied
--- FAIL: TestKeyLabel (0.00s)
```

thaJeztah (approved):
LGTM, thanks!

kolyshkin:
> Flaky test on Fedora
> 
> ```
> === RUN   TestKeyLabel
>     label_linux_test.go:202: write /proc/self/attr/keycreate: permission denied
> --- FAIL: TestKeyLabel (0.00s)
> ```

I've seen this a few times locally and could not understand why. Maybe @rhatdan knows something? Should we add a retry to the code?

kolyshkin (approved):
LGTM, thank you very much!

rhatdan:
Permission denied is either SELinux or DAC.  Would need to see AVC messages to determine which.

Issue #225, opened 2025-03-12
Flake in TestSELinux, TestSocketLabel


kolyshkin:
> Reproduced the same issue with a different test case:
> 
> ```console
> [kir@kir-tp1 go-selinux]$ go test  -run TestSELinux -count 10000 .
> --- FAIL: TestSELinux (0.00s)
>     selinux_linux_test.go:145: system_u:system_r:container_t:s0:c462,c874
>     selinux_linux_test.go:146: system_u:object_r:container_file_t:s0:c462,c874
>     selinux_linux_test.go:148: system_u:system_r:container_t:s0:c43,c985
>     selinux_linux_test.go:149: system_u:object_r:container_file_t:s0:c43,c985
>     selinux_linux_test.go:153: system_u:system_r:container_t:s0:c762,c778
>     selinux_linux_test.go:154: system_u:object_r:container_file_t:s0:c762,c778
>     selinux_linux_test.go:156: ClearLabels
>     selinux_linux_test.go:158: system_u:system_r:container_t:s0:c494,c1001
>     selinux_linux_test.go:159: system_u:object_r:container_file_t:s0:c494,c1001
>     selinux_linux_test.go:163: PID:1817877 MCS:
>     selinux_linux_test.go:168: SetFSCreateLabel failed write /proc/thread-self/attr/fscreate: permission denied
>     selinux_linux_test.go:169: write /proc/thread-self/attr/fscreate: permission denied
> --- FAIL: TestSELinux (0.00s)
>     selinux_linux_test.go:145: system_u:system_r:container_t:s0:c714,c854
>     selinux_linux_test.go:146: system_u:object_r:container_file_t:s0:c714,c854
>     selinux_linux_test.go:148: system_u:system_r:container_t:s0:c152,c966
>     selinux_linux_test.go:149: system_u:object_r:container_file_t:s0:c152,c966
>     selinux_linux_test.go:153: system_u:system_r:container_t:s0:c252,c553
>     selinux_linux_test.go:154: system_u:object_r:container_file_t:s0:c252,c553
>     selinux_linux_test.go:156: ClearLabels
>     selinux_linux_test.go:158: system_u:system_r:container_t:s0:c321,c912
>     selinux_linux_test.go:159: system_u:object_r:container_file_t:s0:c321,c912
>     selinux_linux_test.go:163: PID:1817877 MCS:
[...]

PR #226, opened 2025-03-12
tests: use LockOSThread to avoid flakes

kolyshkin:
The SetFSCreateLabel documentation says the caller should use runtime.LockOSThread. Indeed, if not used, there is an occasional flake:

> $ go test  -run TestSELinux -count 10000 .
> selinux_linux_test.go:168: SetFSCreateLabel failed write /proc/thread-self/attr/fscreate: permission denied
> selinux_linux_test.go:169: write /proc/thread-self/attr/fscreate: permission denied

Add LockOSThread to fix the flake.

While at it, fix logging to avoid printing the same error twice.

Fixes: #225.

### DIFF
--- a/go-selinux/selinux_linux_test.go
+++ b/go-selinux/selinux_linux_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


kolyshkin:
Let me redo it.

kolyshkin:
Redone in #227

Issue #222, opened 2025-03-07
Flaky test: `TestKeyLabel: write /proc/self/attr/keycreate: permission denied`


AkihiroSuda:
> Flaky test on Fedora
> ```
> === RUN   TestKeyLabel
>     label_linux_test.go:202: write /proc/self/attr/keycreate: permission denied
> --- FAIL: TestKeyLabel (0.00s)
> ```

_Originally posted by @AkihiroSuda in https://github.com/opencontainers/selinux/issues/221#issuecomment-2705315717_

This also happens on almalinux 8 (see e.g. https://github.com/opencontainers/selinux/actions/runs/13740062867/job/38428218382#step:7:79).

kolyshkin:
I spent some time looking at it this week, and it seems that
 - it is related to Fedora / latest kernel;
 - it is easy to reproduce

Here's a quick repro:
```console
[kir@kir-tp1 selinux]$ pwd
/home/kir/git/opencontainers/selinux
[kir@kir-tp1 selinux]$ go test -count 1000 -run KeyLabel ./go-selinux/label/
```

I run it on my Fedora 41 laptop and I'm getting frequent failures. If you don't, feel free to increase count or run it again.

I also tried to add retries (into two different places) and it doesn't help. Here's my ugly retrials+debug code: <details>
```diff
diff --git a/go-selinux/selinux_linux.go b/go-selinux/selinux_linux.go
index be139a4..7702596 100644
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -16,6 +16,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/opencontainers/selinux/pkg/pwalkdir"
 	"golang.org/x/sys/unix"
@@ -437,12 +438,22 @@ func writeCon(fpath, val string) error {
 		return err
 	}
 
+	i := 0
+again:
 	if val != "" {
 		_, err = out.Write([]byte(val))
 	} else {
 		_, err = out.Write(nil)
 	}
 	if err != nil {
+		if errors.Is(err, unix.EACCES) {
+			i++
+			if i < 10 {
+				print("writeCon: got EACCES, retry ", i, "\n")
+				time.Sleep(100 * time.Nanosecond)
+				goto again
+			}
+		}
 		return err
 	}
 	return nil
@@ -711,12 +722,21 @@ func peerLabel(fd uintptr) (string, error) {
[...]

PR #223, opened 2025-03-08
CI: add openSUSE Tumbleweed

AkihiroSuda:
Tumbleweed uses SELinux since snapshot 20250211

https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/G3W5NIY3OKRBHPHWTPYEUPSS4LKZN77N/

PR #227, opened 2025-03-12
Improve SetKeyCreate error reporting, fix test flakes

kolyshkin:
Fixes: #222
Fixes: #225

Please review commit by commit. See individual commits for details.

Ideally, I'd like this to be merged after #218 and #223.

### DIFF
--- a/go-selinux/selinux_linux_test.go
+++ b/go-selinux/selinux_linux_test.go
@@ -188,6 +188,12 @@ func TestSocketLabel(t *testing.T) {
 		t.Skip("SELinux not enabled, skipping.")
 	}
 
+	// Ensure the thread stays the same for duration of the test.
+	// Otherwise Go runtime can switch this to a different thread,
+	// which results in EACCES in call to SetSocketLabel.
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
 	label := "system_u:object_r:container_t:s0:c1,c2"
 	if err := SetSocketLabel(label); err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM
Nice work, I wonder if we should do something similar with execcon, which has similar issues.

rhatdan (dismissed):

kolyshkin:
Rebased.

rhatdan:
No I don't think we should change that functionality, just let it go.

rhatdan (approved):

kolyshkin:
@thaJeztah PTAL

rhatdan:
Needs a rebase.

go-selinux/selinux_linux_test.go
```diff
@@ -165,15 +172,13 @@ func TestSELinux(t *testing.T) {
 	if err == nil {
 		t.Log(FSCreateLabel())
 	} else {
-		t.Log("SetFSCreateLabel failed", err)
-		t.Fatal(err)
+		t.Fatal("SetFSCreateLabel failed:", err)
```
thaJeztah, line 175: Nit; `t.Fatal` terminates the test, so we could get rid of the if/else, and instead put the `t.Fatal` in a `if err != nil` branch

thaJeztah (approved):
LGTM, nice!

left one suggestion, but nothing new, so definitely fine for separate

go-selinux/selinux_linux_test.go
kolyshkin: I was just trying to minimize the changes to simplify review. But since you've asked for it... 😁 

Indeed it looks better this way.

kolyshkin:
Rebased; addressed @thaJeztah comment.

thaJeztah (approved):
LGTM, thanks!

kolyshkin:
@rhatdan ptal

rhatdan (approved):

PR #230, opened 2025-03-26
Switch to golangci-lint v2

kolyshkin:
The configuration was migrated using golangci-lint migrate and when
tweaked manually trying to minimize it.

Also, fix some new warnings reported by updated staticcheck.

### DIFF
--- a/go-selinux/label/label_linux.go
+++ b/go-selinux/label/label_linux.go
@@ -18,7 +18,7 @@ var validOptions = map[string]bool{
 	"level":    true,
 }
 
-var ErrIncompatibleLabel = errors.New("Bad SELinux option z and Z can not be used together")
+var ErrIncompatibleLabel = errors.New("bad SELinux option: z and Z can not be used together")
 
 // InitLabels returns the process label and file labels to be used within
 // the container.  A list of options can be passed into this function to alter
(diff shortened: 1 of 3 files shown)


rhatdan (approved):
LGTM

go-selinux/label/label_linux.go
```diff
@@ -18,7 +18,7 @@ var validOptions = map[string]bool{
 	"level":    true,
 }
 
-var ErrIncompatibleLabel = errors.New("Bad SELinux option z and Z can not be used together")
+var ErrIncompatibleLabel = errors.New("bad SELinux option: z and Z can not be used together")
```
thaJeztah, line 21: 👍  Did a quick search to see if there's any code string-matching the error, but I don't think there is (nothing found outside of `vendor/` AFAICS https://grep.app/search?case=true&q=Bad+SELinux+option

go-selinux/label/label_linux.go
```diff
@@ -52,11 +52,11 @@ func InitLabels(options []string) (plabel string, mlabel string, retErr error) {
 				return "", selinux.PrivContainerMountLabel(), nil
 			}
 			if i := strings.Index(opt, ":"); i == -1 {
-				return "", "", fmt.Errorf("Bad label option %q, valid options 'disable' or \n'user, role, level, type, filetype' followed by ':' and a value", opt)
+				return "", "", fmt.Errorf("bad label option %q, valid options 'disable' or \n'user, role, level, type, filetype' followed by ':' and a value", opt)
```
thaJeztah, line 55: Same for this one; https://grep.app/search?case=true&q=Bad+label+option

go-selinux/label/label_linux.go
thaJeztah: Actually; it looks like nobody may be using the `ErrIncompatibleLabel` error as a sentinel error

https://grep.app/search?case=true&q=.ErrIncompatibleLabel


I don't know if there would be any reason for anyone to treat this error special; if there's not, we could consider deprecating it.

(but that's definitely unrelated to this PR!)

thaJeztah (approved):
LGTM (one ready / out of draft)

PR #232, opened 2025-05-20
add SEUserByName, fallback to failsafe context in GetDefaultContextWithLevel

capnspacehook:
Ported getseuserbyname (https://github.com/SELinuxProject/selinux/blob/main/libselinux/src/seusers.c#L188) to Go, and added `failsafe_context` parsing to `GetDefaultContextWithLevel` as that's what libselinux's get_default_context_with_level does (https://github.com/SELinuxProject/selinux/blob/main/libselinux/src/get_context_list.c#L488).

`getseuserbyname` reads the`seusers` file (https://www.man7.org/linux/man-pages/man5/seusers.5.html) to find the SELinux user and the MLS level for a given Linux user.

`get_default_context_with_level` takes a SELinux user, MLS level, and an SELinux context (combination of SELinux user, role, domain, and MLS level, in many cases this context is the context of the caller) and returns the SELinux context that should be used when creating processes as a certain Linux user.

`get_default_context_with_level` attempts to find a suitable context for the specific SELinux user that was passed first, then searches in the list of global context mappings if no match was found, and finally just returns the failsafe context if no match was found anywhere else. `GetDefaultContextWithLevel` previously returned an error if no context could be found for the SELinux user or globally.

Man page for failsafe_context: https://www.man7.org/linux//man-pages/man5/failsafe_context.5.html

### DIFF
--- a/go-selinux/selinux.go
+++ b/go-selinux/selinux.go
@@ -305,11 +305,19 @@ func DisableSecOpt() []string {
 	return []string{"disable"}
 }
 
+// SEUserByName retrieves the SELinux username and security level for a given
+// Linux username. The username and security level is based on the
+// /etc/selinux/{SELINUXTYPE}/seusers file.
+func SEUserByName(username string) (seUser string, level string, err error) {
+	return getSeUserByName(username)
+}
+
 // GetDefaultContextWithLevel gets a single context for the specified SELinux user
 // identity that is reachable from the specified scon context. The context is based
 // on the per-user /etc/selinux/{SELINUXTYPE}/contexts/users/<username> if it exists,
 // and falls back to the global /etc/selinux/{SELINUXTYPE}/contexts/default_contexts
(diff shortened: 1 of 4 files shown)


capnspacehook:
@rhatdan no prob, signed commit

kolyshkin:
@capnspacehook can you split this into two commits, by functionality. Same PR is fine.

kolyshkin (commented):
Regarding the second patch -- I see we're opening two files in `getDefaultContextWithLevel`, while we might not even read the second one. Your patch adds a third file to open.

Maybe we can rework the code first to open the file only when needed (the most changes will be in the tests I guess), then add the failsafe context?

capnspacehook:
@rhatdan done, do you need me to do anything else?

rhatdan:
Rebase and squash your commits, and we can merge.

PR #235, opened 2025-10-07
keyring: fix typo in EACCES check

cyphar:
Commit 965323e01c4e ("SetKeyLabel: add thread group leader requirement")
added verification that the caller of SetKeyLabel is the thread-group
leader, however the check had a typo in it, which would almost always
cause all errors to be treated as ErrNotTGLeader.

It's a bit of a shame that os.Getuid() and os.Getpid() are untyped, as a
one-character typo like this can really easily cause bugs without type
checking...


kolyshkin (approved):
Oh 🤦🏻‍♂️ 

LGTM

rhatdan (approved):
LGTM

PR #237, opened 2025-11-05
selinux: migrate to pathrs-lite procfs API

cyphar:
The previous isProcHandle approach introduced in 03b517d ("selinux:
verify that writes to /proc/... are on procfs") was a fairly naive
solution to CVE-2019-16884 style bugs, as it only checked that the
target was a procfs file without any verification what exact procfs file
it is.

A far more insidious attack (as discussed at the time) would be to
instead bind-mount something like /proc/self/sched on top of
/proc/self/attr/... which would not be detectable using a simple
filesystem type check.

The new pathrs-lite API (provided by filepath-securejoin) can correctly
detect this and includes many other hardenings to avoid attacks of this
kind.

Fixes: [CVE-2025-52881](https://github.com/opencontainers/runc/security/advisories/GHSA-cgrx-mc8f-2prm)

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -17,8 +17,11 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/opencontainers/selinux/pkg/pwalkdir"
+	"github.com/cyphar/filepath-securejoin/pathrs-lite"
+	"github.com/cyphar/filepath-securejoin/pathrs-lite/procfs"
 	"golang.org/x/sys/unix"
+
+	"github.com/opencontainers/selinux/pkg/pwalkdir"
 )
 
 const (
@@ -72,10 +75,6 @@ var (
 	state             = selinuxState{
 		mcsList: make(map[string]bool),
 	}
-
-	// for attrPath()
-	attrPathOnce   sync.Once
-	haveThreadSelf bool
 
 	// for policyRoot()
 	policyRootOnce sync.Once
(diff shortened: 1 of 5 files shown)


go-selinux/selinux_linux.go
```diff
@@ -256,48 +255,187 @@ func readConfig(target string) string {
...
+// "one-shot opens" slightly more ergonomic. The provided mode must be os.O_*
+// flags to indicate what mode the returned file should be opened with (flags
+// like os.O_CREAT and os.O_EXCL are not supported).
+//
+// If no error occurred, the returned handle is guaranteed to be exactly
+// /proc/self/<subpath> with no tricky mounts or symlinks causing you to
+// operate on an unexpected path (with some caveats on pre-openat2 or
+// pre-fsopen kernels).
+//
+// [OpenPid]: https://pkg.go.dev/github.com/cyphar/filepath-securejoin/pathrs-lite/procfs#Handle.OpenPid
```
kolyshkin: Change `[OpenPid]` to `[procfs.Handle.OpenPid]`, and drop this.

kolyshkin (commented):
Minor doc nits, otherwise lgtm

dduzgun-security:
👋, thank you for your work on resolving the issue. To safeguard users relying on this library, would it be feasible to release a patched version prior to any public CVE disclosure in the future please? I think it would reduce scanner alerts (+ pipeline failures) on vulnerabilities with no fixed version and eliminate malicious actors trying to exploit this publicly known vulnerability to cause harm.

cyphar:
I would've preferred that too, but we couldn't pick a later disclosure date (KubeCon is this coming week and I am giving a talk about these security issues at a conference in early December).

Given how often security issues with opencontainers/selinux and runc overlap, perhaps we need to come up with a way that runc maintainers can either do a security release on behalf of opencontainers/selinux or we add more runc maintainers to the set of opencontainers/selinux maintainers (@kolyshkin is a maintainer but I don't know if he has the authority to do releases unilaterally -- @mrunalp and @thaJeztah are also runc maintainers but they weren't involved in this security issue and so might not have been able to help).

@rhatdan What do you think?

rhatdan:
I am fine with adding other maintainers and owners to this Repository.  Planning on Retiring in two weeks or so and taking a big step back.

rhatdan:
LGTM

PR #239, opened 2025-11-08
Enable more golangci-lint

SuperQ:
* Enable additional formatters.
* Enable modernize linter.
* Apply fixes.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,7 +2,20 @@ version: "2"
 
 formatters:
   enable:
+    - gci
     - gofumpt
+    - goimports
+  settings:
+    gci:
+      sections:
+        - standard
+        - default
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/opencontainers/selinux
 
-go 1.19
+go 1.24.0
 
 require (
 	github.com/cyphar/filepath-securejoin v0.6.0
-	golang.org/x/sys v0.26.0
+	golang.org/x/sys v0.37.0
```
thaJeztah: I don't see a real need to update these dependencies; go modules are designed around  MVS (https://research.swtch.com/vgo-mvs), which means library modules should specify the lowest possible version needed. Consumers of the module can decide to use any more current, compatible (according to SemVer) version of the dependencies, including newer versions of Go.

PR #246, opened 2025-11-25
fix: uniqMcs use all cpu

ningmingxiao:
fix   https://github.com/opencontainers/selinux/issues/247

### DIFF
--- a/go-selinux/label/label_linux_test.go
+++ b/go-selinux/label/label_linux_test.go
@@ -53,6 +53,27 @@ func TestInit(t *testing.T) {
 	}
 }
 
+func TestSELinuxLabelExhaustion(t *testing.T) {
+	needSELinux(t)
+	selinux.CategoryRange = 5
+	var testNull []string
+	for i := 0; i < 20; i++ {
+		_, _, err := InitLabels(testNull)
+		wantErr := (i == 19)
+		if wantErr {
+			if err == nil {
+				t.Fatal("err should not be nil")
+			} else if errors.Is(err, selinux.ErrMCSExhausted) {
+				t.Fatalf("unexpected error %s", err.Error())
+			}
+		} else {
+			if err != nil {
+				t.Fatalf("unexpected error in loop %d: %v", i, err)
+			}
+		}
+	}
+}
+
 func TestRelabel(t *testing.T) {
 	needSELinux(t)
 
--- a/go-selinux/selinux.go
+++ b/go-selinux/selinux.go
@@ -20,6 +20,7 @@ const (
 var (
 	// ErrMCSAlreadyExists is returned when trying to allocate a duplicate MCS.
 	ErrMCSAlreadyExists = errors.New("MCS label already exists")
+	ErrMCSExhausted     = errors.New("MCS label exhausted")
 	// ErrEmptyPath is returned when an empty path has been specified.
 	ErrEmptyPath = errors.New("empty path")
 
@@ -212,8 +213,15 @@ func ClearLabels() {
 }
 
 // ReserveLabel reserves the MLS/MCS level component of the specified label
+//
+// Deprecated: Use ReserveLabelV2 instead.
 func ReserveLabel(label string) {
-	reserveLabel(label)
+	_ = reserveLabel(label)
+}
+
+// ReserveLabelV2 reserves the MLS/MCS level component of the specified label returns an error on failure
+func ReserveLabelV2(label string) error {
+	return reserveLabel(label)
 }
 
 // CheckLabel check the MLS/MCS level component of the specified label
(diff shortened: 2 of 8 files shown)


kolyshkin (commented):
Maybe we need to change the public API to add error reporting to, say, selinux.ContainerLabels (and similar functions using addMcs/uniqMcs)?
Keep the old functions for backward compatibility but deprecate those?

rhatdan:
Needs a rebase.  @ningmingxiao still working on this one?

ningmingxiao:
done thanks  @rhatdan

kolyshkin (changes requested):
So, what we need to do here is move selinux.ContainerLabels (and selinux.ContainerLabelsSize) into an internal package. It looks like no one is using `selinux.ContainerLabels` -- the only user is `label.InitLabels. Means we can move it to internal and then do whatever we want to. See https://github.com/opencontainers/selinux/issues/247#issuecomment-4309556073 for details.

ningmingxiao:
done  @kolyshkin

go-selinux/label/label_linux_test.go
```diff
@@ -53,6 +54,22 @@ func TestInit(t *testing.T) {
 	}
 }
 
+func TestSELinuxLabelExhaustion(t *testing.T) {
+	needSELinux(t)
+	selinux.CategoryRange = 5
+	var testNull []string
+	for i := 0; i < 20; i++ {
+		_, _, err := InitLabels(testNull)
+		if i == 19 {
```
kolyshkin: So we don't check err unless `i == 19`?

I'd suggest something like
```go
wantErr := (i == 19)
if wantErr {
  ...
} else {
   ...
}
```

go-selinux/label/label_linux_test.go
```diff
@@ -53,6 +54,22 @@ func TestInit(t *testing.T) {
...
+			if err == nil {
+				t.Fatal("err should not be nil")
+			} else if !strings.Contains(err.Error(), "SELinux label exhaustion") {
```
kolyshkin: Should you do `if !errors.Is(err, ErrMCSExhausted)`?

go-selinux/selinux_linux.go
```diff
@@ -38,11 +38,13 @@ const (
 	xattrNameSelinux = "security.selinux"
 )
 
+var maxSelinuxLabelSize = int(CategoryRange * (CategoryRange - 1) / 2)
```
kolyshkin, line 41: Alas, `CategoryRange` can be (and is) changed by any other package.

I'm fixing this issue in https://github.com/opencontainers/selinux/pull/262 by introducing `SetCategoryRange` and deprecating `CategoryRange`.

go-selinux/selinux.go
```diff
@@ -217,8 +218,15 @@ func ClearLabels() {
 }
 
 // ReserveLabel reserves the MLS/MCS level component of the specified label
+//
+// Deprecated: Use ReserveLabelV2 instead.
 func ReserveLabel(label string) {
```
kolyshkin, line 218: Do we have any external users?

[Yes we do; plenty.](https://sourcegraph.com/search?q=context%3Aglobal+selinux.ReserveLabel&patternType=keyword&sm=0&__cc=1&df=%5B%22lang%22%2C%22Go%22%2C%22lang%3Ago%22%5D&df=%5B%22file%22%2C%22Exclude+vendor%22%2C%22-file%3A%28%5E%7C%2F%29vendor%2F%22%5D)

go-selinux/selinux.go
```diff
@@ -217,8 +218,15 @@ func ClearLabels() {
...
-	reserveLabel(label)
+	_ = reserveLabel(label)
+}
+
+// ReserveLabelV2 reserves the MLS/MCS level component of the specified label
```
kolyshkin: nit: document the errors returned.

go-selinux/selinux_linux.go
```diff
@@ -835,21 +843,26 @@ func clearLabels() {
...
+func reserveLabel(label string) error {
 	if len(label) != 0 {
 		con := strings.SplitN(label, ":", 4)
 		if len(con) > 3 {
-			_ = mcsAdd(con[3])
+			err := mcsAdd(con[3])
+			if errors.Is(err, ErrMCSExhausted) {
+				return err
+			}
+			return nil
```
kolyshkin: Why is any other error ignored?

ningmingxiao: I afraid when user try to reserve the label user have to ignore the exist error, I'm not sure whether to return it  @kolyshkin

go-selinux/selinux_linux.go
ningmingxiao: you can merge 262 into main I will rebase it @kolyshkin

go-selinux/selinux_linux.go
ningmingxiao: I already fix it.

PR #248, opened 2025-12-04
When mcs are exhausted, uniqMcs should return an error

zzzzzzzzzy9:
There should be an exit branch in the infinite loop in the code, and return error should be inside the for loop.
Additionally, when judging the length of mcsList, the delete method of mcs should directly remove the key-value pair instead of setting the key-value pair to false, otherwise the length of mcsList will keep increasing.

### DIFF
--- a/go-selinux/label/label_linux.go
+++ b/go-selinux/label/label_linux.go
@@ -30,7 +30,10 @@ func InitLabels(options []string) (plabel string, mlabel string, retErr error) {
 	if !selinux.GetEnabled() {
 		return "", "", nil
 	}
-	processLabel, mountLabel := selinux.ContainerLabels()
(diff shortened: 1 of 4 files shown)


zzzzzzzzzy9:
https://github.com/opencontainers/selinux/issues/247
https://github.com/opencontainers/selinux/pull/246

kolyshkin (changes requested):
NACK. You change the public API in an incompatible way, it is not acceptable

rhatdan:
@zzzzzzzzzy9 are you still interested in adding a new API?

kolyshkin:
I guess we should come up with the new API and deprecate the old one; then remove the old one ~1 year later.

rhatdan:
Needs a rebase.

PR #251, opened 2026-04-06
fix infinite loop in SEUserByName

capnspacehook:
The previous logic would loop forever if the last line of the `seusers` file was empty or only contained whitespace, as the check on 1337 would be true. I opted to fix the issue by using a `bufio.Scanner` as using it is simpler and less error prone.

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -1326,18 +1326,13 @@ func getSeUserFromReader(username string, gids []string, r io.Reader, lookupGrou
 	var groupSeUser, groupLevel string
 
 	lineNum := -1
-	reader := bufio.NewReader(r)
-	for {
-		lineBytes, readErr := reader.ReadBytes('\n')
-		if readErr != nil {
-			if !errors.Is(readErr, io.EOF) {
-				return "", "", fmt.Errorf("failed to read seusers file: %w", readErr)
-			}
-		}
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		rawLine := scanner.Text()
 		lineNum++
 
 		// remove any trailing comments, then extra whitespace
-		line, _, _ := strings.Cut(string(lineBytes), "#")
+		line, _, _ := strings.Cut(rawLine, "#")
 		line = strings.TrimSpace(line)
 		if line == "" {
 			continue
@@ -1377,10 +1372,9 @@ func getSeUserFromReader(username string, gids []string, r io.Reader, lookupGrou
 			defaultSeUser = seUserField
 			defaultLevel = levelField
 		}
-
-		if errors.Is(readErr, io.EOF) {
-			break
-		}
+	}
+	if err := scanner.Err(); err != nil {
+		return "", "", fmt.Errorf("failed to read seusers file: %w", err)
 	}
 
 	if groupSeUser != "" {
--- a/go-selinux/selinux_linux_test.go
+++ b/go-selinux/selinux_linux_test.go
@@ -619,14 +619,15 @@ func TestGetSeUser(t *testing.T) {
 			seUserBuf: `
 system_u:system_u:s0-s15:c0.c255
 root:root:s0-s15:c0.c255
-bob:staff_u:s0-s15:c0.c255`,
+bob:staff_u:s0-s15:c0.c255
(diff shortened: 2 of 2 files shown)


rhatdan:
Please fix your commit message to pass the validation step.

capnspacehook:
@rhatdan done

rhatdan:
LGTM

@kolyshkin @thaJeztah PTAL

thaJeztah:
Will have a look; I have a feeling that we may not need to switch to a scanner, but need to have a look on my machine

go-selinux/selinux_linux.go
```diff
@@ -1326,18 +1326,13 @@ func getSeUserFromReader(username string, gids []string, r io.Reader, lookupGrou
...
+	for scanner.Scan() {
+		rawLine := scanner.Text()
 		lineNum++
 
 		// remove any trailing comments, then extra whitespace
-		line, _, _ := strings.Cut(string(lineBytes), "#")
+		line, _, _ := strings.Cut(rawLine, "#")
 		line = strings.TrimSpace(line)
 		if line == "" {
 			continue
```
thaJeztah, line 1338: Instead of changing the code to a scanner;  I think the problem can be fixed by handling the `io.EOF` here and breaking out of the loop instead of continuing;

```suggestion
		if line == "" {
			if errors.Is(readErr, io.EOF) {
				break
			}
			continue
```

The essence here is that `bufio.Reader.ReadBytes` returns BOTH an error (which can be `io.EOF`) _and_ the bytes that were read. If it's a `io.EOF` we should process the data (if any) then break the loop; https://pkg.go.dev/bufio#Reader.ReadBytes

> ... it returns the data read before the error and the error itself (often io.EOF).

thaJeztah: I'm overall OK with changing to a scanner, but wasn't sure if there's other changes that would introduce (ISTR it uses a default buffer-size that would be different); @kolyshkin any preference?

capnspacehook: If there's a specific reason that `bufio.Scanner` is not preferable here it would make sense to continue to use `bufio.Reader`; otherwise I strongly lean towards using `bufio.Scanner`. It's what I originally used when I opened #232 because it's easier to use (and harder to misuse) imo, and I don't know of any significant downsides.

Your suggestion would fix the issue I believe, but I thought it would be better to take the opportunity to switch to the `bufio.Scanner` which makes it harder to make mistakes like this in the first place.

kolyshkin: `bufio.NewReader` uses 4K buffer, `bufio.NewScanner` uses dynamic buffer, initially sized to 4K and it can grow for up to 64K.

I agree we can fix a bug in two ways. The rest of the code uses Scanner so I slightly gravitate towards it. 

PS white at it, I'd like to reduce the number of allocated strings in the code by doing something like this:
```diff
               // remove any trailing comments, then extra whitespace
-               line, _, _ := strings.Cut(string(lineBytes), "#")
-               line = strings.TrimSpace(line)
+               lineBytes, _, _ = bytes.Cut(lineBytes), "#")
+               lineBytes = bytes.TrimSpace(line)
```
and overall using bytes for as long as possible (when converting the result to string). Otherwise we have too many garbage to collect (since string is immutable).

capnspacehook: That makes sense, TIL the bytes package had those functions too

capnspacehook: @kolyshkin are you OK with sticking with `bufui.Scanner`, because if so I wouldn't need to switch to the functions in the `bytes` package

kolyshkin: OK, let's not overdo things here, the current solution LGTM

kolyshkin (approved):
LGTM

thaJeztah (approved):
LGTM

Sorry for the extra discussion, I recalled @kolyshkin mentioned not using a scanner in https://github.com/opencontainers/selinux/pull/232#discussion_r2414874852, so wanted to be sure there wasn't a specific reason for that 👍

capnspacehook:
@thaJeztah no worries, you were just doing your due diligence 👍 thanks for the review!

PR #253, opened 2026-04-17
selinux_linux: fix path in execLabel() (regression in v1.13.0)

mschiff:
execLabel() was trying to read `/proc/thread-self/exec` instead of `/proc/thread-self/attr/exec`

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -564,7 +564,7 @@ func pidLabel(pid int) (string, error) {
 // ExecLabel returns the SELinux label that the kernel will use for any programs
 // that are executed by the current process thread, or an error.
 func execLabel() (string, error) {
-	return readConThreadSelf("exec")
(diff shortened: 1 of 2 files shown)


kolyshkin:
@mschiff thank you! The fix is correct (and the issue is introduced in commit c8cfa6f). Unfortunately it was missed during a review, and we don't have a test case to catch that.

I have force-pushed to your branch to fix the missing Signed-off-by and add a second commit with the test case.

The test fails before the fix like this:
```
=== RUN   TestExecLabel
    selinux_linux_test.go:236: open /proc/thread-self/exec handle: unsafe procfs detected: openat2 /proc/thread-self/exec: no such file or directory
--- FAIL: TestExecLabel (0.00s)
```

kolyshkin (approved):

kolyshkin:
@thaJeztah PTAL

thaJeztah (approved):
LGTM, thx!

PR #255, opened 2026-04-18
Assorted CI bumps and related fixes

kolyshkin:
See individual commits for details. Nothing major.

PR #256, opened 2026-04-20
Switch to go 1.22; modernize code

kolyshkin:
Go 1.22 was released more than 2 years ago (in February 2024) and is not supported since Go 1.24 is out (February 2025).

I think we can switch to it as a minimally required version, and apply some new language and stdlib features.

See individual commits for details.

~~Draft until #255 is merged.~~

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -83,26 +83,15 @@ const (
 
 var (
 	readOnlyFileLabel string
-	state             = selinuxState{
+
+	state = selinuxState{
 		mcsList: make(map[string]bool),
 	}
-
-	// for policyRoot()
-	policyRootOnce sync.Once
-	policyRootVal  string
-
-	// for label()
-	loadLabelsOnce sync.Once
-	labels         map[string]string
 )
 
-func policyRoot() string {
-	policyRootOnce.Do(func() {
-		policyRootVal = filepath.Join(selinuxDir, readConfig(selinuxTypeTag))
-	})
-
-	return policyRootVal
-}
+var policyRoot = sync.OnceValue(func() string {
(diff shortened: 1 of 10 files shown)


go-selinux/selinux_linux.go
```diff
@@ -40,11 +40,9 @@ const (
 )
 
 type selinuxState struct {
-	mcsList       map[string]bool
-	selinuxfs     string
-	selinuxfsOnce sync.Once
-	enabledSet    bool
-	enabled       bool
+	mcsList    map[string]bool
```
thaJeztah: The only concern I have is that we're splitting state; we have a package-level `selinuxState` that holds state, and now we're splitting the `getSelinuxMountPoint` to be a separate state.

While _probably_ there's no difference in practice, it does make it an odd design; we should either use package level vars for all state, or use a single(ton) for the state.


Perhaps good to move this commit separate if we don't want to block the go1.22 update.

kolyshkin: I'm not sure why the path to selinux mount is part of the state; it shouldn't be.

Looking into git history, it was added by the initial commit 34bec73, so can't say why it was implemented this way.

Anyway, separated out to #258

thaJeztah: Yeah, I think I have looked at some of that in the past as well. I think it'd be good to revisit whether we need the struct at all, and if we do (e.g. for testing to prevent a global state?) to make it more useful.

Currently, I don't think it offers much over package level vars.

thaJeztah (approved):
LGTM

thaJeztah:
@cyphar @rhatdan ptal 🤗

rhatdan (approved):
LGTM

PR #257, opened 2026-04-22
Use math/rand/v2 rather than crypto/rand

### LINKED ISSUE #252: Build tag to not include `crypto/rand`
Would you consider the inclusion of a PR that adds a build tag (tentative name `opencontainers_selinux.nocrypto`, perhaps) that elides `uniqMcs` and its call tree from the build? Our usecase is a minimal binary to spawn shells for users (including transitioning to the user's context via `SetExecLabel`, if selinux is enabled), and since this is the last source of inclusion of any package in `crypto/...`, getting rid of it would mean not having to care about FIPS compliance at all (as well as losing us a tiny bit of binary size, but that's a very minor thing).

espadolini:
Since consumers of the library might want to completely avoid using crypto/... to ensure compliance with FIPS 140-3, and since using slightly more predictable MCS labels shouldn't have any effect on security, this commit replaces the one use of crypto/rand in the codebase with its equivalent in math/rand/v2.

Closes #252

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -3,13 +3,12 @@ package selinux
 import (
 	"bufio"
 	"bytes"
-	"crypto/rand"
-	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
 	"io/fs"
 	"math/big"
+	"math/rand/v2"
 	"os"
 	"os/user"
 	"path/filepath"
@@ -954,16 +953,16 @@ func mcsDelete(mcs string) {
 
 func uniqMcs(catRange uint32) string {
 	var (
-		n      uint32
 		c1, c2 uint32
 		mcs    string
 	)
 
 	for {
-		_ = binary.Read(rand.Reader, binary.LittleEndian, &n)
-		c1 = n % catRange
-		_ = binary.Read(rand.Reader, binary.LittleEndian, &n)
-		c2 = n % catRange
+		//#nosec G404 -- using slightly more predictable MCS labels won't affect security, so it's fine to use math/rand/v2 here.
+		{
+			c1 = rand.Uint32N(catRange)
+			c2 = rand.Uint32N(catRange)
+		}
 		if c1 == c2 {
 			continue
 		} else if c1 > c2 {
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,6 +8,7 @@ linters:
   enable:
     - copyloopvar   # Detects places where loop variables are copied.
     - dupword       # Detects duplicate words.
+    - depguard      # Allows or rejects certain package imports.
     - errorlint     # Detects code that may cause problems with Go 1.13 error wrapping.
     - gocritic      # Metalinter; detects bugs, performance, and styling issues.
     - gosec         # Detects security problems.
@@ -22,6 +23,16 @@ linters:
     - unconvert     # Detects unnecessary type conversions.
     - usetesting    # Reports uses of functions with replacement inside the testing package.
   settings:
+    depguard:
+      rules:
+        fipscompliance:
+          deny:
+            - pkg: crypto/
+              desc: Using crypto/... requires consumers to potentially have to care about FIPS compliance, so it should be avoided if possible.
+        mathrandv1:
+          deny:
+            - pkg: math/rand$
+              desc: Use math/rand/v2.
     govet:
       enable-all: true
       settings:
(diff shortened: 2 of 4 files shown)


go-selinux/selinux_linux.go
```diff
@@ -3,13 +3,12 @@ package selinux
...
 	"bufio"
 	"bytes"
-	"crypto/rand"
-	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
 	"io/fs"
 	"math/big"
+	"math/rand"
```
thaJeztah: Should we use `math/rand/v2` ? https://pkg.go.dev/math/rand/v2

espadolini: go.mod specifies `go 1.19` and the test matrix also tests with go 1.19, but `math/rand/v2` was added in 1.22. We could switch between the two based on build constraints, but is that worth it?

thaJeztah: Doh! I forgot for a bit that it required go1.22, but good news is that  we have a PR pending to update the minimum version, so if that one gets merged, that would allow switching here;

- https://github.com/opencontainers/selinux/pull/256

thaJeztah:
Looks like the linter complains, but possibly switching to v2 would fix that as well;

```
 Error: go-selinux/selinux_linux.go:1007:8: G404: Use of weak random number generator (math/rand or math/rand/v2 instead of crypto/rand) (gosec)
  		c1 = rand.Uint32() % catRange
  		     ^
  Error: go-selinux/selinux_linux.go:1008:8: G404: Use of weak random number generator (math/rand or math/rand/v2 instead of crypto/rand) (gosec)
  		c2 = rand.Uint32() % catRange
  		     ^
```

espadolini:
Using math/rand/v2 there would also fail the linter, unfortunately. I've added a nosec directive for those two uses specifically (in 6452e76c717e794f55899ad83be41ba106db97f8) rather than removing the G404 gosec rule altogether so any future use of math/rand or math/rand/v2 will also require a justification.

go-selinux/selinux_linux.go
```diff
@@ -1000,16 +999,17 @@ func intToMcs(id int, catRange uint32) string {
...
 		c1, c2 uint32
 		mcs    string
 	)
 
 	for {
-		_ = binary.Read(rand.Reader, binary.LittleEndian, &n)
-		c1 = n % catRange
-		_ = binary.Read(rand.Reader, binary.LittleEndian, &n)
-		c2 = n % catRange
+		// #nosec G404 - using slightly more predictable MCS labels won't affect
```
kolyshkin: nit: comments targeted for non-humans (e.g. linters) should have no space after `//`. Frankly I don't remember where it's coming from but all the existing #nosec comments follow the rule, and so should you

thaJeztah: Yeah, they're directives; https://go.dev/doc/comment#directives (similar to `//go:build`)

But `#nosec` has always been an odd one, because even without a space it's not following the standard, because tool names can only exist of alphanumeric characters, so `#` is not allowed.

thaJeztah:
> Using math/rand/v2 there would also fail the linter, unfortunately

Ah, yes, it was there in plain sight; 😂 

```
(math/rand or math/rand/v2 instead of crypto/rand) (gosec)
```

go-selinux/selinux_linux.go
espadolini: Fixed the formatting to match the other nosec comments in b09086dc4933028d14d752ef5443a1d9401b82cd

thaJeztah: oh! could you squash the commits?


It's also fine to keep that for later; I think the go1.22 bump should be merged soon (waiting for a second approval); in that case, I think we should opt for using `math/v2` to keep things fresh 😅

kolyshkin: @espadolini you can actually switch to math/rand/v2 right now (CI on go 1.19 will fail for now but the rebase after the merge will fix it).

go-selinux/selinux_linux.go
rhatdan: That PR is merged, so let's update this PR

go-selinux/selinux_linux.go
espadolini: Rebased and updated to use math/rand/v2; I also went ahead and added a depguard lint against math/rand v1 since there's really no reason to use it if v2 is available and it was a very small change to begin with - let me know if I should undo that or maybe split it in a different commit.

rhatdan (approved):
LGTM

kolyshkin (approved):
lgtm

kolyshkin:
@espadolini thanks for depguard additions and fixes in `pkg/pwalk*`!

PR #258, opened 2026-04-22
Simplify getSelinuxMountPoint

kolyshkin:
Use sync.OnceValue, and remove selinuxfs from struct state.

No change in functionality.

Draft until #256 is merged.

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -39,11 +39,9 @@ const (
 )
 
 type selinuxState struct {
-	mcsList       map[string]bool
-	selinuxfs     string
-	selinuxfsOnce sync.Once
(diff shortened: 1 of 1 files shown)


kolyshkin:
The aim of this PR is to simplify the code.

There is a [concern about separating the state](https://github.com/opencontainers/selinux/pull/256#discussion_r3124300558), but frankly I don't get it; the selinux mount point is not related to the rest of the `struct selinuxState` members. The way it is implemented now just wraps the code into an extra level of indirection without adding any value.

rhatdan (approved):
LGTM

kolyshkin:
@thaJeztah ptal

thaJeztah (approved):
LGTM

PR #259, opened 2026-04-22
Remove intToMcs

kolyshkin:
Since commit b03a082 ("Rework SELinux bindings to match Go Interface Standards") this function is private and is only used by a single test.

I ran the test and it did not show anything for MCS:

    selinux_linux_test.go:338: PID:3113385 MCS:

This is because PID is too high.

Most probably it was an idea to derive unique MCS categories from the container ID or something like that, but apparently that idea was ditched.

Let's ditch the code, too.

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -978,26 +978,6 @@ func mcsDelete(mcs string) {
 	state.mcsList[mcs] = false
 }
 
-func intToMcs(id int, catRange uint32) string {
-	var (
-		SETSIZE = int(catRange)
-		TIER    = SETSIZE
-		ORD     = id
-	)
-
-	if id < 1 || id > 523776 {
-		return ""
-	}
-
-	for ORD > TIER {
-		ORD -= TIER
-		TIER--
-	}
-	TIER = SETSIZE - TIER
-	ORD += TIER
-	return fmt.Sprintf("s0:c%d,c%d", TIER, ORD)
-}
-
 func uniqMcs(catRange uint32) string {
 	var (
 		n      uint32
--- a/go-selinux/selinux_linux_test.go
+++ b/go-selinux/selinux_linux_test.go
@@ -314,8 +314,6 @@ func TestSELinux(t *testing.T) {
 	t.Log(flabel)
 	ReleaseLabel(plabel)
 
-	pid := os.Getpid()
-	t.Logf("PID:%d MCS:%s", pid, intToMcs(pid, 1023))
 	err = SetFSCreateLabel("unconfined_u:unconfined_r:unconfined_t:s0")
 	if err != nil {
 		t.Fatal("SetFSCreateLabel failed:", err)


kolyshkin:
I noticed the issue when reading the sources, as no linter flags it.

Yet I later found a way for a linter to flag it:
```console
[kir@kir-tp1 selinux]$ golangci-lint run --tests=false ./...
go-selinux/selinux_linux.go:975:6: func intToMcs is unused (unused)
func intToMcs(id int, catRange uint32) string {
     ^
1 issues:
* unused: 1
```

Added such thing to https://github.com/opencontainers/selinux/pull/260.

thaJeztah:
> Most probably it was an idea to derive unique MCS categories from the container ID or something like that, but apparently that idea was ditche

Yup! I went down the rabbit-hole once to look things up for this; https://github.com/opencontainers/selinux/issues/247#issuecomment-3589858124

> > I wonder if we could just remove the arbitrary limits entirely, but I'm sure there's a good reason for them.
>
> I was curious about the limit as well, and earlier this week tried to trace back where the 1024 limit originated from, which needed some tracking across multiple repositories, because this code moved around;
>
> * it started as a package in docker/docker (moby)
> * was moved to libcontainer in the docker/docker (moby) repository
> * which became runc
> * then got moved to this repository to become a separate module
>
> In either case, the original 1024 limit looks to have been added in [moby/moby@4c43566](https://github.com/moby/moby/commit/4c4356692580afb3971094e322aea64abe0e2500) as part of this PR;
>
> * [Patch to add SELinux support to docker. moby/moby#4211](https://github.com/moby/moby/pull/4211)
>
> https://github.com/moby/moby/blob/4c4356692580afb3971094e322aea64abe0e2500/pkg/selinux/selinux.go#L356 https://github.com/moby/moby/blob/4c4356692580afb3971094e322aea64abe0e2500/pkg/selinux/selinux.go#L248-L266
>
> At a first glance, I couldn't find discussion around that limit, so it could indeed have been just an arbitrary limit
>
> I did find a later change in this repository that would allow setting a custom range through the exported `CategoryRange` variable (although using a package-var for that feels a bit clunky);
>
> * [Allow the category range to be changed #100](https://github.com/opencontainers/selinux/pull/100)
>
> > Allow the category range to be changed
> > This keeps the range as part of the global state of the package but allows users to modify the upper bounds of the category range.


So, yes, it was originally added in https://github.com/moby/moby/commit/4c4356692580afb3971094e322aea64abe0e2500 (https://github.com/moby/moby/pull/4211) to get a unique ID, but later replaced by `uniqMcs` in https://github.com/moby/moby/commit/b7942ec2ca7c7568df0c3b7eb554b05e2c3a3081 (https://github.com/moby/moby/pull/5448)

> The daemon will now generate a new unique random SELinux label when a
> container starts, and remove it when the container is removed.   The MCS
> labels will be stored in the daemon memory.  The labels of containers will
> be stored in the container.json file.
>
> When the daemon restarts on boot or if done by an admin, it will read all containers json files and reserve the MCS labels.
>
> A potential problem would be conflicts if you setup thousands of containers,
> current scheme would handle ~500,000 containers.

thaJeztah (approved):
LGTM

thaJeztah:
tracing back history is fun for these, because

- it was originally in moby (docker/docker)
- then moved to the libcontainer _package_ in moby; https://github.com/moby/moby/commit/6158ccad97db51e756aafefb096d1163aa4d6439 (https://github.com/moby/moby/pull/6317)
- which became the `libcontainer` repository; https://github.com/docker-archive/libcontainer
- which then became `runc`  https://github.com/opencontainers/runc
- and then the selinux code was extracted to this repository

rhatdan (approved):
LGTM

PR #261, opened 2026-04-24
Simplify/remove some code

kolyshkin:
A few removals/simplifications. See individual commits for details.


kolyshkin (commented):
Technically I can also remove calls to `SetFSCreateLabel`, `FSCreateLabel`, `CurrentLabel` and `ExecLabel` from selinux_stub_test.go (as they are no longer in stubs). But it can stay.

Same applies for other functions that were previously removed from stubs in commit c0c5de3. They all test the same thing (stub implementations of readConThreadSelf/writeConThreadSelf). Again, they can stay (although I don't see lot of value).

thaJeztah (approved):
LGTM

rhatdan:
LGTM

PR #262, opened 2026-04-25
Deprecate some stuff

kolyshkin:
It turns out we need to change the API to allow some functions return errors (to fix #247).

To do that, we need to first:
 - deprecate the existing API;
 - introduce the new API.
 
 This is what this PR does. The next step (not done here yet) is major refactoring (see https://github.com/opencontainers/selinux/issues/247#issuecomment-4309556073).
 
 Please review commit by commit, see commit descriptions for more info.

### DIFF
--- a/go-selinux/label/label_linux.go
+++ b/go-selinux/label/label_linux.go
@@ -30,7 +30,7 @@ func InitLabels(options []string) (plabel string, mlabel string, retErr error) {
 	if !selinux.GetEnabled() {
 		return "", "", nil
 	}
-	processLabel, mountLabel := selinux.ContainerLabels()
+	processLabel, mountLabel := selinux.ContainerLabels() //nolint:staticcheck // ContainerLabels will be moved to an internal package.
 	if processLabel == "" {
 		// processLabel is required; if empty, do nothing.
 		return processLabel, mountLabel, nil
--- a/go-selinux/label/label_linux.go
+++ b/go-selinux/label/label_linux.go
@@ -71,7 +71,9 @@ func InitLabels(options []string) (plabel string, mlabel string, retErr error) {
 		if pcon["level"] != mcsLevel {
 			selinux.ReleaseLabel(processLabel)
 		}
-		selinux.ReserveLabel(p)
+		if err := selinux.ReserveLabelV2(p); err != nil {
+			return "", "", err
+		}
 		processLabel = p
 	}
 	mountLabel = mcon.Get()
--- a/go-selinux/label/label_linux_test.go
+++ b/go-selinux/label/label_linux_test.go
@@ -24,9 +24,6 @@ func TestInit(t *testing.T) {
 		t.Fatalf("InitLabels failed: %v:", err)
 	}
 	testDisabled := []string{"disable"}
-	if selinux.ROFileLabel() == "" {
-		t.Fatal("selinux.ROFileLabel: empty")
-	}
 	plabel, mlabel, err := InitLabels(testDisabled)
 	if err != nil {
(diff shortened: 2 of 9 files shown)


thaJeztah:
I saw you mentioned "internal" in one of the other comments; what if we'd;


1. move the whole `labels/` package to `internal/labels` (this would show up as a lateral move, so all history is preserved nicely)
2. make the `labels/` package wrappers around the internal implementation (probably needs a second commit to not confuse git)
3. ☝️ the internal implementation can already have the implementation that has an error return (https://github.com/opencontainers/selinux/issues/247), but the exported `labels/xxx` wrapper would discard the error.

thaJeztah:
Oh! I guess the error-returns are breaking changes in selinux, not the labels package? 🤔 

I guess something similar could still apply; have the implementations internal, then have the exported packages be shallow wrappers around the internal ones.

FWIW; I think indeed many of these may not be used externally, but we need to double-check; I once made a similar move where a deprecated function was wrapped in another code-base, so removing a deprecated wrapper broke that (even though the function was not called anywhere) 😞

kolyshkin:
@thaJeztah I already have a set of patches that move all of selinux implementation into an internal package, so that both selinux and label can access it. Will open a PR soon.

This one is a prerequisite to the above; it marks some stuff we'll have to remove later, and introduces the alternative/better API. Can you please review?

I can tell some more about these changes:

1. We have `CategoryRange` as a public variable. Any user of the package can change its value, and it affects the behavior of *ContainerLabels functions. Alas, if we want to move the implementation of *ContainerLabels to an internal package, we can't leave the variable as-is (without introducing circular dependencies).

2. Deprecate `ROFileLabel` (to be removed). I could not find any existing users ([using this](https://sourcegraph.com/search?q=context%3Aglobal+selinux.ROFileLabel&patternType=keyword&sm=0&df=%5B%22file%22%2C%22Exclude+vendor%22%2C%22-file%3A%28%5E%7C%2F%29vendor%2F%22%5D&__cc=1)) and its implementation is rather questionable as the results depend on whether `ContainerLabels` were called first. Having said that, there is a way to fix the implementation (as suggested in the commit message), but I'd rather not revive something that is not used.

3. Deprecate `ContainerLabels` (to be removed). Same thing -- [no external users](https://sourcegraph.com/search?q=context%3Aglobal+selinux.ROFileLabel&patternType=keyword&sm=0&df=%5B%22file%22%2C%22Exclude+vendor%22%2C%22-file%3A%28%5E%7C%2F%29vendor%2F%22%5D&__cc=1); everyone is using `label.InitLabels` instead.

4. Deprecate `{KVM,Init}ContainerLabels` (to be removed) and introduce a replacement API with error return added (`{KVM,Init}ContainerLabel`). This is to fix the uniqMcs issue.

Now, all the calls that are being removed have a suggestion to open an issue in case those are still used. The plan is to:

 - cut a release with all the above;
 - wait for feedback and adapt in case we missed something;
 - make another release with the deprecated stuff removed, and the fix for uniqMcs.
 
Now, I have the "move to internal" mostly ready but it's a very massive change. So I'd like to get this one agreed upon first.

rhatdan (dismissed):
LGTM

kolyshkin:
We also need ReserveLabel which returns an error.

Added.

kolyshkin:
Updated SetCategoryLabel to error out on values > 1024 (which appears to be the kernel hard limit).

rhatdan:
LGTM Again.
@thaJeztah PTAL

thaJeztah (approved):
LGTM

PR #263, opened 2026-04-27
selinux: release map key when call mcsDelete

ningmingxiao:
this pr may reduce the memory, old version never delete the map it may cause caontainerd/docker use more unnecessary memory when try to create and delete container many times.

### DIFF
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -39,7 +39,7 @@ const (
 )
 
 type selinuxState struct {
-	mcsList    map[string]bool
+	mcsList    map[string]struct{}
 	enabledSet bool
 	enabled    bool
 	sync.Mutex
@@ -82,7 +82,7 @@ var (
 	readOnlyFileLabel string
 
 	state = selinuxState{
-		mcsList: make(map[string]bool),
+		mcsList: make(map[string]struct{}),
 	}
 )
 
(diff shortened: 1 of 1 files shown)


rhatdan:
Don't see this saving a lot, but LGTM.
@kolyshkin @thaJeztah PTAL

go-selinux/selinux_linux.go
```diff
@@ -924,7 +924,7 @@ func mcsDelete(mcs string) {
 	}
 	state.Lock()
 	defer state.Unlock()
-	state.mcsList[mcs] = false
+	delete(state.mcsList, mcs)
```
thaJeztah, line 910: If we're changing the logic to be based on whether the key is present (not on the value), we should change the type;

```go
type selinuxState struct {
	mcsList    map[string]struct{}
```

That saves allocations and makes it clear that the value is irrelevant, and prevents introducing accidental bugs (if someone changes code to set the value to `false`).

ningmingxiao: @rhatdan  done thanks
If the caller is daemon process it may use unnecessary memory.

ningmingxiao: thanks @rhatdan  @thaJeztah

thaJeztah (approved):
LGTM

kolyshkin (approved):
LGTM

PR #264, opened 2026-04-28
[DO NOT MERGE] move selinux implementation to internal package

kolyshkin:
~~Draft pending #262 merge.~~

The issue being solved is go-selinux is both a package for external consumption, as well as a package to be used by  go-selinux/label. Some of go-selinux APIs (say ContainerLabels) have only 1 user: label, and it needs to be internal.

The main idea is we can
 - use the functionality freely from both packages (go-selinux and go-selinux/label);
 - add or change the (internal) API however we want.
 
 TODO: identify more stuff that should be internal-only (like ContainerLabels).

----

1. go-selinux: move to internal/impl
    
    This moves (git mv) the whole package to under internal, with the
    minimal changes (package name only).
    
    This commit breaks git-bisect; this is done deliberately to show this is
    merely the moving files around. The next commit fixes things.

2. go-selinux: wire to ../internal/impl
    
    The only change in internal/impl is making ReadConThreadSelf and
    WriteConThreadSelf public.

3. label: switch to internal/impl

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,2 +1,2 @@
 [codespell]
-skip = ./.git,./go.sum,./go-selinux/testdata
+skip = ./.git,./go.sum,,./internal/impl/testdata
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -147,7 +147,7 @@ jobs:
       # https://github.com/opencontainers/selinux/issues/225
       - name: "racy test"
         continue-on-error: true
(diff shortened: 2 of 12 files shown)


rhatdan:
Other then hating the `impl` name this LGTM.

kolyshkin:
> Other then hating the `impl` name this LGTM.

Well I don't like it either; any suggestions? `sys`?

PR #267, opened 2026-05-06
Fix label.InitLabels regression in v1.14.0; amend ReserveLabelV2 doc

kolyshkin:
1. ci: fix ignoring test failures on lima
    
    Commit c826d9e introduced a regression -- if make test is failed, the job succeeds.
    
    The fix is to
     - not use continue-on-error (which makes the test succeed);
     - use "if: always()" for the step we want to run even on failure.

2. label: fix tests
    
    The two tests (TestInit and TestFileLabel) try to reserve the same MCS
    label (`level:s0:c1,c15`) without releasing it, and so the second test
    fails with ErrMCSAlreadyExists (as it should, since we are using
    selinux.ReserveLabelV2 which can return errors):
    
    > === RUN   TestFileLabel
    >     label_linux_test.go:122: InitLabels(user) failed: MCS label already exists
    
    Fix tests by releasing the label.

3. label: fix ErrMCSAlreadyExists from InitLabels
    
    The hardening done to label.InitLabels in commit 97d2a7c ("Add
    ReserveLabelV2, deprecate ReserveLabel") revealed the internal issue:
    we try to reserve label which was already reserved. As a result,
    InitLabels can return ErrMCSAlreadyExists when it should not.
    
    Fix by only reserving a label if its MCS ("level") differs from the one
    returned by ContainerLabels.
    
    Add a test (which fails before the fix).

### DIFF
--- a/go-selinux/label/label_linux.go
+++ b/go-selinux/label/label_linux.go
@@ -22,8 +22,13 @@ var ErrIncompatibleLabel = errors.New("bad SELinux option: z and Z can not be us
 
 // InitLabels returns the process label and file labels to be used within
 // the container.  A list of options can be passed into this function to alter
-// the labels.  The labels returned will include a random MCS String, that is
-// guaranteed to be unique.
+// the labels.
+//
+// Unless the "level" option is provided (to set a custom level), the labels
+// returned will include a random MCS string guaranteed to be unique in the
+// scope of the process using this package. If the "level" option is provided,
+// the custom level set is reserved but not checked to be unique.
+//
 // If the disabled flag is passed in, the process label will not be set, but the mount label will be set
 // to the container_file label with the maximum category. This label is not usable by any confined label.
 func InitLabels(options []string) (plabel string, mlabel string, retErr error) {
@@ -70,9 +75,11 @@ func InitLabels(options []string) (plabel string, mlabel string, retErr error) {
 	if p := pcon.Get(); p != processLabel {
 		if pcon["level"] != mcsLevel {
 			selinux.ReleaseLabel(processLabel)
-		}
-		if err := selinux.ReserveLabelV2(p); err != nil {
-			return "", "", err
+			// Ignore ErrMCSAlreadyExists as label is user-specified and might be
+			// already reserved (e.g. when containers in a pod use the same label).
+			if err := selinux.ReserveLabelV2(p); err != nil && !errors.Is(err, selinux.ErrMCSAlreadyExists) {
+				return "", "", err
+			}
 		}
 		processLabel = p
 	}
--- a/go-selinux/label/label_linux_test.go
+++ b/go-selinux/label/label_linux_test.go
@@ -127,3 +127,15 @@ func TestFileLabel(t *testing.T) {
 		t.Fatalf("InitLabels(filetype) failed: %v", err)
 	}
 }
+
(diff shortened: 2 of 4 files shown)


rhatdan:
LGTM

kolyshkin:
Draft until I test this with podman (v1.4.0 fails frequently with ErrMCSAlreadyExists).

go-selinux/label/label_linux_test.go
```diff
@@ -43,6 +43,7 @@ func TestInit(t *testing.T) {
 	if plabel != "user_u:user_r:user_t:s0:c1,c15" || (mlabel != "user_u:object_r:container_file_t:s0:c1,c15" && mlabel != "user_u:object_r:svirt_sandbox_file_t:s0:c1,c15") {
 		t.Fatalf("InitLabels(user) failed (plabel=%q, mlabel=%q)", plabel, mlabel)
 	}
+	selinux.ReleaseLabel(plabel)
```
Luap99: should the release not be above the test assert, otherwise if the check fails you do not release and cause cascading failures?

kolyshkin: Thought about this, too. Normally we don't expect this to fail but yes, if it fails it's better to not cause further failures.

Fixed.

go-selinux/label/label_linux.go
```diff
@@ -70,7 +70,10 @@ func InitLabels(options []string) (plabel string, mlabel string, retErr error) {
 	if p := pcon.Get(); p != processLabel {
 		if pcon["level"] != mcsLevel {
 			selinux.ReleaseLabel(processLabel)
-			if err := selinux.ReserveLabelV2(p); err != nil {
+			err := selinux.ReserveLabelV2(p)
+			// Ignore ErrMCSAlreadyExists as label is user-specified and might be
+			// already reserved (e.g. when containers in a pod using the same label).
+			if err != nil && !errors.Is(err, selinux.ErrMCSAlreadyExists) {
```
thaJeztah: Perhaps keep it scoped? Also, should this be squashed with the previous commit?

```go
if err := selinux.ReserveLabelV2(p); err != nil && !errors.Is(err, selinux.ErrMCSAlreadyExists) {
```

thaJeztah (dismissed):
LGTM; left one nit

go-selinux/label/label_linux.go
kolyshkin: Switched to scoped `err` (split because the line was too long).

Squashed with the previous commit (they solve 2 different issues but with the same outcome, and touch the same code so why not).

kolyshkin: Also, updated InitLabels documentation to better explain when the labels returned contains unique MCS (only if within the same process (like dockerd or containerd) and if it's not overridden). Maybe podman takes care of reserving the known/used labels, I dunno.

thaJeztah:
Looks like we need to set some sensible timeouts on the actions 😞 VM task looks to got stuck, so probably will continue running for 6 hours by default

Screenshot 2026-05-08 at 10 55 27

rhatdan (approved):
LGTM

kolyshkin:
> Looks like we need to set some sensible timeouts on the actions 😞 VM task looks to got stuck, so probably will continue running for 6 hours by default

Right; see https://github.com/opencontainers/selinux/pull/270

PR #269, opened 2026-05-08
label.InitLabels: optimize

kolyshkin:
A few shortcuts for specific usage scenarios:

1. When there are no options specified, label.InitLabels is equivalent to selinux.ContainerLabels. Return its result with no further processing.

2. When the only option specified is "disable", don't call selinux.ContainerLabels as the call is somewhat expensive and we won't use its results (moreover, we have to free the just-generated MCS).

We still allow "disable" to be set together with other options for backward compatibility, but this should eventually be forbidden.

Also, previously "disable:<value>" option was silently ignored, now it results in an error (as it should be).

### DIFF
--- a/go-selinux/label/label_linux.go
+++ b/go-selinux/label/label_linux.go
@@ -10,7 +10,6 @@ import (
 
 // Valid Label Options
 var validOptions = map[string]bool{
-	"disable":  true,
 	"type":     true,
 	"filetype": true,
 	"user":     true,
@@ -35,9 +34,13 @@ func InitLabels(options []string) (plabel string, mlabel string, retErr error) {
 	if !selinux.GetEnabled() {
 		return "", "", nil
 	}
+	if len(options) > 0 && options[0] == "disable" {
+		return "", selinux.PrivContainerMountLabel(), nil
+	}
 	processLabel, mountLabel := selinux.ContainerLabels() //nolint:staticcheck // ContainerLabels will be moved to an internal package.
-	if processLabel == "" {
-		// processLabel is required; if empty, do nothing.
+	if processLabel == "" || len(options) == 0 {
+		// 1. processLabel is required; if empty, do nothing.
+		// 2. If there are no options to process, we're done.
 		return processLabel, mountLabel, nil
 	}
 	defer func() {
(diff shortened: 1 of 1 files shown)


rhatdan:
Needs a rebase

kolyshkin:
Rebased; PTAL @thaJeztah

rhatdan (approved):
LGTM

thaJeztah (approved):
LGTM

thaJeztah:
Is this blocking the merge? Wondering if we should disable that requirement; it rarely really works;

Screenshot 2026-05-12 at 19 56 38

kolyshkin:
> Is this blocking the merge? Wondering if we should disable that requirement; it rarely really works;

It works by ensuring the CI run for a PR is done against the latest code. Once a year or so I am seeing a case when CI is green, but when a PR is merged everything is broken. In short, this is why I enabled this setting.

kolyshkin:
> > Is this blocking the merge? Wondering if we should disable that requirement; it rarely really works;
> 
> It works by ensuring the CI run for a PR is done against the latest code. Once a year or so I am seeing a case when CI is green, but when a PR is merged everything is broken. In short, this is why I enabled this setting.

We also have a "update via rebase" button (unfortunately I don't know how to disable the "update via merge") that can be used to rebase (and re-run CI against current HEAD).

thaJeztah:
Yeah, it's really rare to happen though, and probably using a merge queue could help with that?

kolyshkin:
> Yeah, it's really rare to happen though, and probably using a merge queue could help with that?

Sorry for my ignorance; what is a merge queue?

thaJeztah:
github allows stacking merges instead of merging directly; I _think_ that also does a merge (and running CI) before doing the actual merge;

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/merging-a-pull-request-with-a-merge-queue
 
> The merge queue provides the same benefits as the Require branches to be up to date before merging branch protection, but does not require a pull request author to update their pull request branch and wait for status checks to finish before trying to merge.

PR #271, opened 2026-05-12
Add SetProcessKind

### LINKED ISSUE #266: RFC: introduce SetProcessKind
OK, I found out a somewhat weird thing (which I wish to find earlier than today). NOTE this is a continuation of https://github.com/opencontainers/selinux/issues/247#issuecomment-4309556073

All (open source) users of selinux.KVMContainerLabel[s] and selinux.InitContainerLabel[s] are immediately releasing the acquired MCS. They only use the `type` field. Here are all the users:
 * containerd: https://github.com/containerd/containerd/blob/main/internal/cri/seutil/seutil.go
 * podman: https://github.com/containers/podman/blob/main/pkg/selinux/selinux.go
 * cri-o: https://github.com/cri-o/cri-o/blob/main/server/selinux.go

What happens is we generate uniq MCS which gets released immediately! The only used field is `type`.

So I think here in selinux we don't need KVMContainerLabel[s], InitContainerLabel[s] etc.

We only need something like this:

```go
type ContainerLabelType int

const (
  ContainerProcess ContainerLabelKind = iota // Corresponds to "process" label.
  ContainerInitProcess // ... "init_process" label.
  ContainerKVMProcess // ... "kvm_process" label.
)

func (k ContainerLabelType) configKey() (primary, fallback string, ok bool) {                                                                                                                                                 
      switch k {  
      case ContainerProcess:                                                                                                                                                                                                    
          return "process", "", true
      case ContainerInitProcess:
          return "init_process", "process", true
      case ContainerKVMProcess:                                                                                                                                                                                                 
          return "kvm_process", "process", true
      }                                                                                                                                                                                                                         
      return "", "", false
}

func ChangeContainerLabelType(cLabel string, type ContainerLabelType) (string, error) {
      if cLabel == "" {                                                                                                                                                                                                         
          return "", nil
      }                                                                                                                                                                                                                         
      primary, fallback, ok := kind.configKey()
      if !ok {                                                                                                                                                                                                                  
          return "", fmt.Errorf("selinux: invalid ContainerLabelKind %d", kind)
      }                                                                                                                                                                                                                         
                                                                                                                                                                                                                                
      src := label(primary)
      if src == "" && fallback != "" {                                                                                                                                                                                          
          src = label(fallback)
      }                                                                                                                                                                                                                         
      if src == "" {
          return cLabel, nil                                                                                                                                                                                                    
      }           

      srcCtx, err := newContext(src)
      if err != nil {
          return "", err                                                                                                                                                                                                        
      }
      dstCtx, err := newContext(cLabel)                                                                                                                                                                                         
      if err != nil {
          return "", err
      }
      dstCtx["type"] = srcCtx["type"]
      return dstCtx.Get(), nil 
}
```

Then the users are simplified to (using podman as an example)
```go
// KVMLabel returns labels for running kvm isolated containers
func KVMLabel(cLabel string) (string, error) {
   return selinux.ChangeContainerLabelType(cLabel, selinux.ContainerKVMProcess)
}

// InitLabel returns labels for running systemd based containers.
func InitLabel(cLabel string) (string, error) {
   return selinux.ChangeContainerLabelType(cLabel, selinux.ContainerInitProcess)
}
```

cri-o and containerd change is similar. Most importantly, this will remove unneeded and expensive "generate and reserve MCS labels to immediately discard those".

@rhatdan @thaJeztah what do you think?

kolyshkin:
All (open source) users of `KVMContainerLabel[s]` and `InitContainerLabel[s]` (containerd, podman, and cri-o) are immediately releasing the acquired MCS. They only need the type field, so they can change the existing label to that type. Everything else, including the just-generated unique MCS label, is not used.

Introduce `SetProcessKind` which does just what all those users need.

Fixes: #266 

Testing this:
 - podman: https://github.com/containers/podman/pull/28693
 - cri-o: https://github.com/cri-o/cri-o/pull/9955
 - containerd: https://github.com/containerd/containerd/pull/13395

### DIFF
--- a/go-selinux/selinux.go
+++ b/go-selinux/selinux.go
@@ -48,6 +48,21 @@ var (
 	privContainerMountLabel string
 )
 
+// ProcessKind selects which process domain [SetProcessKind] applies to a label.
+type ProcessKind int
+
+const (
+	ProcessKindRegular ProcessKind = 1
+	ProcessKindInit    ProcessKind = 2
+	ProcessKindKVM     ProcessKind = 3
+)
+
+// SetProcessKind returns label with its type component replaced by the one
+// corresponding to kind. Other label components are kept intact.
+func SetProcessKind(label string, kind ProcessKind) (string, error) {
+	return setProcessKind(label, kind)
+}
+
 // Context is a representation of the SELinux label broken into 4 parts
 type Context map[string]string
 
@@ -292,6 +307,8 @@ func KVMContainerLabels() (string, string) {
 
 // KVMContainerLabel returns the default process label to be used
 // for KVM containers by the calling process.
+//
+// If you only need to change a type of existing label, use [SetProcessKind] instead.
 func KVMContainerLabel() (string, error) {
 	return kvmContainerLabel()
 }
@@ -306,6 +323,8 @@ func InitContainerLabels() (string, string) {
 
 // InitContainerLabel returns the default process label to be used
 // for containers running an init system like systemd by the calling process.
+//
+// If you only need to change a type of existing label, use [SetProcessKind] instead.
 func InitContainerLabel() (string, error) {
 	return initContainerLabel()
 }
--- a/go-selinux/selinux_linux.go
+++ b/go-selinux/selinux_linux.go
@@ -1513,3 +1513,46 @@ func getDefaultContextWithLevel(user, level, scon string) (string, error) {
 
 	return getDefaultContextFromReaders(&c)
 }
+
+func (k ProcessKind) keys() (primary, fallback string, ok bool) {
+	switch k {
+	case ProcessKindRegular:
+		return "process", "", true
+	case ProcessKindInit:
+		return "init_process", "process", true
+	case ProcessKindKVM:
+		return "kvm_process", "process", true
+	}
+	return "", "", false
+}
+
+func setProcessKind(cLabel string, k ProcessKind) (string, error) {
+	if cLabel == "" {
+		return "", nil
+	}
+	primary, fallback, ok := k.keys()
+	if !ok {
+		return "", fmt.Errorf("selinux.SetProcessKind: invalid ProcessKind %d", k)
+	}
+
+	src := label(primary)
+	if src == "" && fallback != "" {
+		src = label(fallback)
+	}
+	if src == "" {
+		return cLabel, nil
+	}
+
+	// Replace cLabel type with one from src.
+	srcCtx, err := newContext(src)
+	if err != nil {
+		return "", fmt.Errorf("selinux.SetProcessKind: invalid %s label %s: %w", primary, src, err)
+	}
+	dstCtx, err := newContext(cLabel)
+	if err != nil {
+		return "", fmt.Errorf("selinux.SetProcessKind: invalid label %s: %w", cLabel, err)
+	}
+
+	dstCtx["type"] = srcCtx["type"]
+	return dstCtx.get(), nil
+}
--- a/go-selinux/selinux_linux_test.go
+++ b/go-selinux/selinux_linux_test.go
@@ -84,6 +84,69 @@ func TestSetFileLabel(t *testing.T) {
 	}
 }
 
+func TestSetProcessKind(t *testing.T) {
+	// These cases do not depend on SELinux being enabled.
+	t.Run("empty label", func(t *testing.T) {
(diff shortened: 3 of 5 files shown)


go-selinux/selinux.go
```diff
@@ -48,6 +48,22 @@ var (
 	privContainerMountLabel string
 )
 
+// LabelType is a type to use for [ChangeLabelType]
+type LabelType int
+
+const (
+	TypeProcess LabelType = iota + 1
```
thaJeztah: I'm always a bit cautious on using iota for public API; go makes it easy to still allow untyped integers to be used (or to be cast to the right type), so if the order would ever change, that could break uses, and wouldn't be immediately obvious here.

kolyshkin: Do you want to add a test case to check the order hasn't changed? Or would you prefer to not use `iota` at all?

kolyshkin:
Testing this:
 - podman: https://github.com/containers/podman/pull/28693
 - cri-o: https://github.com/cri-o/cri-o/pull/9955
 - containerd: https://github.com/containerd/containerd/pull/13395

go-selinux/selinux.go
thaJeztah: Perhaps no `iota` at all, and just explicitly set the values; although I don't expect it to happen in this repository, I've been bit a few times by cases where values were implicitly changed (e.g. some refactor sorting consts, or someone adding a new consts somewhere in between).

While _technically_ we could consider the consts to be the contract, I have seen situations where values were either hard-coded, or (not sure if that can happen for this one) persisted, so updating the values would (sometimes _subtly_) break those cases.

Explicitly setting values reduces the temptation to update values as it's more clear that changing them means you're changing the API.


(Hope I make sense 😅)

kolyshkin: You do make sense; yet I still like iota because it's concise. I think we can prevent the values being changed by adding a test case.

Well, after adding a test case it's no so concise (if we count the test as well).

Switched to numbers.

kolyshkin: @thaJeztah ptal

go-selinux/selinux.go
```diff
@@ -48,6 +48,22 @@ var (
...
 )
 
+// LabelType is a type to use for [ChangeLabelType]
+type LabelType int
+
+const (
+	TypeProcess     = LabelType(1)
+	TypeInitProcess = LabelType(2)
+	TypeKVMProcess  = LabelType(3)
+)
```
thaJeztah, line 58: I think this is more idiomatic / common;

```suggestion
const (
	TypeProcess     LabelType = 1
	TypeInitProcess LabelType = 2
	TypeKVMProcess  LabelType = 3
)
```

kolyshkin: fixed

thaJeztah (dismissed):
LGTM

go-selinux/selinux.go
```diff
@@ -48,6 +48,22 @@ var (
...
 )
 
+// LabelType is a type to use for [ChangeLabelType]
+type LabelType int
+
+const (
+	TypeProcess     LabelType = 1
+	TypeInitProcess LabelType = 2
+	TypeKVMProcess  LabelType = 3
+)
```
kolyshkin: Maybe change `Type` to `Kind`?

go-selinux/selinux.go
```diff
@@ -48,6 +48,22 @@ var (
 	privContainerMountLabel string
 )
 
+// LabelType is a type to use for [ChangeLabelType]
```
kolyshkin: nit: missing `.`

go-selinux/selinux.go
kolyshkin: Changed
 * ChangeLabelType → SetProcessKind
 * LabelType → ProcessKind
 * Type{Process,InitProcess,KVMProcess} → ProcessKind{Regular,Init,KVM}

thaJeztah (approved):
code changes LGTM


for naming (naming is hard!) I'd like to defer to more SELinux "native" folks to decide what's the most natural term / language to use. @rhatdan ?

rhatdan:
The kind versus type here is not really an SELinux thing.  These are move Containers/VM Labels that leaked into this repository.  So the upstream SELinux people would have no opinion.  

LGTM

rhatdan (approved):

